PR #2, opened 2019-07-25
Add race condition tests to dendrite unit tests

anoadragon453:
cc @cnly 

Enables the race condition checker when running Dendrite's unit tests.

The other half of this PR (that enables the checks on Sytest) is here: https://github.com/matrix-org/dendrite/pull/745/ Though currently this makes sytest runs quite slow, so whether we want to merge that before fixing the races is questionable.

I don't consider this PR blocked on that one.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -35,14 +35,14 @@ steps:
           limit: 3
 
   - command:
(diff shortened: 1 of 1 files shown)


richvdh:
if this isn't being worked on, I'll close it.

PR #3, opened 2019-08-28
sydent: Add support for matrix_is_tester

dbkr:
<del>nb. This is pinned to a different branch and should be updated to point to master once https://github.com/matrix-org/matrix-is-tester/pull/5/ is merged.</del> Now updated

### DIFF
--- a/sydent/pipeline.yml
+++ b/sydent/pipeline.yml
@@ -9,7 +9,7 @@ steps:
       - "echo '--- Install dependencies'"
       - "pip install -e ."
       - "echo '--- Install matrix-is-tester'"
-      - "python -m pip install git+https://github.com/matrix-org/matrix-is-tester.git@dbkr/bind_denied_2"
+      - "python -m pip install git+https://github.com/matrix-org/matrix-is-tester.git"
       - "echo '+++ Run Tests'"
       - "PYTHONPATH=`pwd` trial matrix-is-tester"
     label: "matrix-is-tester"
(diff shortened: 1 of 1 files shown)


sydent/pipeline.yml
```diff
@@ -5,3 +5,18 @@ steps:
     plugins:
       - docker#v3.0.1:
           image: "matrixdotorg/sydent-build"
+  - command:
+      - "echo '--- Install dependencies'"
+      - "pip install -e ."
+      - "echo '--- Install matrix-is-tester'"
+      - "python -m pip install git+https://github.com/matrix-org/matrix-is-tester.git"
```
richvdh: why do we `pip install` above and `python -m pip install` here?

sydent/pipeline.yml
```diff
@@ -5,3 +5,18 @@ steps:
...
+      - "echo '+++ Run Tests'"
+      - "PYTHONPATH=`pwd` trial matrix-is-tester"
```
richvdh: what's the point of setting PYTHONPATH if we've already pip installed it?

richvdh (commented):
Couple of questions.

> nb. This is pinned to a different branch and should be updated to point to master once matrix-org/matrix-is-tester#5 is merged.

this is outdated?

sydent/pipeline.yml
dbkr: uh, probably just copied from somewhere

sydent/pipeline.yml
dbkr: mm, good point

richvdh (approved):
lgtm. Would be nice to see it running somewhere but maybe that's tricky before it lands?

dbkr:
Bleh, I wrote 4 words of a reply on this and then forgot about it for a week. But yes, we could fudge all the branches to get it running and then change them all across the various projects once we merge things - I'm not sure it's that worthwhile in this case.

PR #4, opened 2019-09-11
Move 'label' to start of each job, for clarity

richvdh:


richvdh:
I'm assuming that there are no objections to this.

PR #5, opened 2019-09-11
Upload the trial logs as artifacts

richvdh:


richvdh:
assuming there are no objections here.

PR #8, opened 2019-10-03
Add annotation uploading to dendrite

anoadragon453:
Will add a nice display to Dendrite's Sytest results similar to what Synapse currently has:

image

~~Blocked on having an `annotate.md` file generated by the Dendrite sytest script.~~ (A Perl implementation is provided by https://github.com/matrix-org/sytest/pull/714).

Here is where Synapse's is generated: https://github.com/matrix-org/sytest/blob/f686445b98d375a1f060bd8e2775b05656b18193/docker/synapse_sytest.sh#L89-L92


richvdh (approved):
lgtm

PR #9, opened 2019-10-04
Add CI jobs for synapse_port_db

babolivier:
Paired with https://github.com/matrix-org/synapse/pull/6140

Defines the Synapse CI jobs to run the tox environment added in https://github.com/matrix-org/synapse/pull/6140

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -332,7 +332,7 @@ steps:
       TRIAL_FLAGS: "-j 8"
     command:
       - "python -m pip install tox"
-      - "tox -e synapse_port_db"
+      - "bash .buildkite/scripts/test_synapse_port_db.sh"
     plugins:
       - docker-compose#v2.1.0:
           run: testenv
@@ -357,7 +357,7 @@ steps:
       TRIAL_FLAGS: "-j 8"
     command:
       - "python -m pip install tox"
-      - "tox -e synapse_port_db"
+      - "bash .buildkite/scripts/test_synapse_port_db.sh"
     plugins:
       - docker-compose#v2.1.0:
           run: testenv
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -325,6 +325,56 @@ steps:
...
+      - artifacts#v1.2.0:
+          upload: [ "_trial_temp/*/*.log" ]
+#      - matrix-org/coveralls#v1.0:
+#          parallel: "true"
+    retry:
+      automatic:
+        - exit_status: -1
+          limit: 2
+        - exit_status: 2
+          limit: 2
```
richvdh: what is the thinking behind this config?

synapse/pipeline.yml
```diff
@@ -325,6 +325,56 @@ steps:
         - exit_status: 2
           limit: 2
 
+  - label: "synapse_port_db / :python: 3.5 / :postgres: 9.5"
+    agents:
+      queue: "medium"
+    env:
+      TRIAL_FLAGS: "-j 8"
```
richvdh: unless the test uses `trial` (I don't think it does?) there is no point in this, and no point in the artifact upload pattern below.

richvdh (dismissed):

synapse/pipeline.yml
```diff
@@ -325,6 +325,37 @@ steps:
...
+    plugins:
+      - docker-compose#v2.1.0:
+          run: testenv
+          config:
+            - .buildkite/docker-compose.py35.pg95.yaml
+      - artifacts#v1.2.0:
+          upload: [ "_trial_temp/*/*.log" ]
+#      - matrix-org/coveralls#v1.0:
+#          parallel: "true"
+
```
richvdh, line 342:
```suggestion
```

richvdh (dismissed):
lgtm, but needs to wait for matrix-org/synapse#6140

richvdh (approved):

PR #13, opened 2019-11-08
move synapse's docker-compose files out-of-tree

richvdh:
... because they are too powerful to load from an untrusted location.


michaelkaye (approved):
This looks... reasonable to me

PR #15, opened 2019-11-21
Have sytest CI use same-named branch of Synapse/Dendrite if exists, else use the dev branch

anoadragon453:
It annoys me that sytest runs keep failing because they're using Synapse develop instead of the branch their intended to pair with. Hopefully this'll also prevent some accidental bugs in the future.

### DIFF
--- a/sytest/pipeline.yml
+++ b/sytest/pipeline.yml
@@ -4,7 +4,7 @@ steps:
       queue: "medium"
     command:
       - "mkdir -p /src"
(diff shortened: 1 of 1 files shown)


richvdh (approved):
looks fine.

This should probably be a script in the sytest repo, which should give some more info about what's going on. Indeed, we presumably already have something similar for checking out the right sytest branch for synapse?

anyway, this is fine for now.

PR #16, opened 2019-11-25
Mount the sytest working copy read-only

richvdh:
... to stop people doing things that will make it crap over my working copy

cf https://github.com/matrix-org/sytest/pull/748


richvdh:
at this point, I'm assuming nobody cares.

PR #18, opened 2019-12-03
enable replication torturing for synapse worker mode

richvdh:
this helps find race conditions in the worker-mode tests.


erikjohnston (approved):
How much time does this add to the test runs ooi?

richvdh:
:man_shrugging:

richvdh:
ftr, it doesn't seem to make much difference to the total runtime. It certainly makes things a bunch flakier though. I'm going to revert this for now; will revisit it one day.

PR #19, opened 2019-12-03
Add sygnal pipeline

anoadragon453:
Coupled with https://github.com/matrix-org/sygnal/pull/74

### DIFF
--- a/sygnal/pipeline.yml
+++ b/sygnal/pipeline.yml
@@ -1,0 +1,27 @@
+steps:
+  - command:
+      - "python -m pip install black"
+      - "black --check ."
+    label: "Check Code Formatting"
+    plugins:
+      - docker#v3.0.1:
+          image: "python:3.7"
+
+  - command:
+      - "python -m pip install flake8"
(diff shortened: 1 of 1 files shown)


sygnal/pipeline.yml
```diff
@@ -0,0 +1,27 @@
...
+      - docker#v3.0.1:
+          image: "python:3.7"
+
+  - command:
+      - "python -m pip install flake8"
+      - "flake8 ."
+    label: "Check Code Style"
+    plugins:
+      - docker#v3.0.1:
+          image: "python:3.7"
```
richvdh, line 16: in an ideal world, these would probably be combined into one pipeline, with a `lint.sh` in the sytest repo (since the linting steps should be kept in step with the code rather than being defined separately in the pipeline repo).

Still, that can happen another time.

richvdh (approved):
lgtm.

@anoadragon453 : as a follow-up PR, can you set `mount-buildkite-agent: false` in each of the docker plugin configs?

PR #22, opened 2019-12-18
Set mount-buildkite-agent: false for all docker tasks on matrix-appservice-irc

Half-Shot:
As discussed, to enable community PRs.


richvdh (approved):
lgtm

PR #23, opened 2019-12-18
Use propagate-uid-gid for irc bridge pipeline

Half-Shot:
Because otherwise npm cannot build.


matrix-appservice-irc/pipeline.yml
```diff
@@ -15,16 +15,19 @@ steps:
     plugins:
       - docker#v3.0.1:
           image: "node:10"
+          propagate-uid-gid: true # So NPM won't complain about permission issues
           mount-buildkite-agent: false
-
```
richvdh, line 19: why are we removing this?

richvdh (approved):

PR #24, opened 2019-12-20
Run worker sytests with split out databases

erikjohnston:
Enables the feature added in https://github.com/matrix-org/sytest/pull/775, but should be backwards compat as we still specify `POSTGRES` env var


richvdh (approved):
m-m-m-m-multi postgres.

PR #27, opened 2020-01-06
Add towncrier and postgres tests to the irc-bridge pipeline

Half-Shot:
Which have been taken from https://github.com/matrix-org/matrix-appservice-irc/blob/03c1943530e2915140f6806572587f557f2917c2/.buildkite/pipeline.yml


matrix-appservice-irc/pipeline.yml
```diff
@@ -34,3 +44,14 @@ steps:
       - docker#v3.0.1:
           image: "node:12"
           mount-buildkite-agent: false
+
+  - label: ":newspaper: Newsfile"
+    branches: "!master !develop !release-*"
+    command:
+      - "python3 -m pip install towncrier"
```
richvdh: I'd recommend sticking these commands in a script (or tox.ini) in the target repo that can be run from here, to reduce the chances of having to PR more changes against the pipeline.

matrix-appservice-irc/docker-compose.postgres.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.1'
+
+services:
+  postgres:
+    image: postgres:11
```
richvdh, line 5: fwiw the synapse tests add

```
    command: -c fsync=off
```

to speed things up marginally.

matrix-appservice-irc/pipeline.yml
Half-Shot: This is done in https://github.com/matrix-org/matrix-appservice-irc/pull/990

richvdh (approved):
lgtm

PR #28, opened 2020-01-08
Run sydent unit tests

anoadragon453:
We have unit tests for Sydent now \o/ So we should probably have CI run them.

### DIFF
--- a/sydent/pipeline.yml
+++ b/sydent/pipeline.yml
@@ -1,10 +1,16 @@
 steps:
   - command:
-    - "/sydent-ci.sh"
-    label: "Check Sydent starts"
+      - "echo '--- Install dependencies'"
+      - "pip install -e ."
+      - "echo '--- Run unit tests'"
+      - "trial tests"
+    label: ":partyparrot: Unit tests"
     plugins:
       - docker#v3.0.1:
           image: "matrixdotorg/sydent-build"
+    artifact_paths:
+      - "_trial_temp/test.log"
+
   - command:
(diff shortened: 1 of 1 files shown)


sydent/pipeline.yml
```diff
@@ -1,18 +1,24 @@
 steps:
   - command:
-    - "/sydent-ci.sh"
-    label: "Check Sydent starts"
+      - "echo '--- Install dependencies'"
```
richvdh, line 3: might be better to update `sydent-ci.sh` instead?

richvdh (dismissed):

sydent/pipeline.yml
anoadragon453: Thing is I don't actually know where that script is... I think it may've been removed and it's still sitting in the latest generated docker image.

anoadragon453: The script exists in the closed PR that also brought about this docker image. https://github.com/matrix-org/sydent/pull/193/files#diff-c2bd4c3d264a03f6f7582437f460d552

So it never became part of the repo.

richvdh: fair enough.

I still think it would be a good idea for these lists of commands to be in the `sydent` git repo; or (in this case) maybe make `tox` work and this can then be

>      - "python -m pip install tox"
>      - "tox -e py27"
 
or something.

still, I guess we're going to be iterating here so this is an improvement for now.

richvdh (approved):

PR #29, opened 2020-01-09
Run some of the synapse tests against python 3.8/postgres 12

richvdh:


synapse/pipeline.yml
```diff
@@ -183,7 +189,7 @@ steps:
         - exit_status: 2
           limit: 2
 
-  - label: ":python: 3.7 / :postgres: 9.5"
+  - label: ":python: 3.7 / :postgres: 11"
```
erikjohnston, line 192: Why this change? Why are we testing pg11?

erikjohnston (approved):
Its fine, what's the rationale for the other bits of shuffling ooi?

synapse/pipeline.yml
richvdh: Github isn't doing the best job of showing a sensible diff.

I was thinking it would be sensible to test against a range of postgres versions rather than just min and max, hence I'm suggesting we retain the current tests against py3.7/pg11.

Rather, I'm replacing the tests against  py3.7/pg9.5 with tests against py3.8/pg12; the rationale being that we already testing against pg9.5 (in the py3.5/pg9.5 tests), and I'm not super-keen to introduce even more test runs.

erikjohnston: Ooooh, right yes. Agreed.

PR #31, opened 2020-01-24
We use yarn now for bifrost

Half-Shot:
we use yarn for bifrost now, which means if we try to CI with npm then we won't use the yarn.lock file, which isn't a great test


michaelkaye (approved):
This looks like a safe change if we've moved to yarn.

PR #33, opened 2020-01-29
Add jobs for Python 3 for Sydent

babolivier:
Add CI jobs for testing Python 3 for Sydent. These jobs are soft-failed, since these jobs will fail until all of [the Python 3 compatibility PRs](https://github.com/matrix-org/sydent/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3Apy3) are merged.

### DIFF
--- a/sydent/pipeline.yml
+++ b/sydent/pipeline.yml
@@ -2,9 +2,9 @@ steps:
   - command:
       - "echo '--- Install dependencies'"
       - "pip install -e ."
-      - "echo '--- Run unit tests'"
+      - "echo '+++ Run unit tests'"
       - "trial tests"
-    label: ":partyparrot: Unit tests"
+    label: ":partyparrot: Unit tests / :python: 2.7"
     plugins:
       - docker#v3.0.1:
           image: "python:2.7"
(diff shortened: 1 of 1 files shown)


michaelkaye (approved):
I think this looks OK

sydent/pipeline.yml
```diff
@@ -18,11 +18,83 @@ steps:
...
           image: "python:2.7"
     artifact_paths:
       - "matrix_is_test/sydent.stderr"
       - "_trial_temp/test.log"
+
+  - command:
+      - "echo '--- Install dependencies'"
+      - "pip install -e ."
+      - "echo '+++ Run unit tests'"
+      - "trial tests"
```
richvdh, line 34: it'd sure be swell to stick these 4 lines in a script so that we don't need to have 3 copies of them here. Likewise the ones for the IS tester.

richvdh:
well it's kinda fine as long as we don't turn on contributor builds for sydent: it doesn't have the magic "make-that-a-safe-thing-to-do" settings, which means having the pipeline in this repo at all is a bit pointless.

ototh, it's not a new thing, so sure.

PR #38, opened 2020-02-13
Run long CI steps lint, sytest in parallel

anoadragon453:
Linting takes quite a while, and so does running the sytests. The overall CI pipeline would be faster if these steps were run in parallel.

Conversely though, we don't want to run them and gum up CI if your code doesn't even build. So run that step initially, along with unit tests as both are very quick.

To be clear, this changes the pipline from:

- Lint
- -- wait --
- Build
- Unit Tests
- SyTests

to

- Build
- Unit Tests
- -- wait --
- Lint
- SyTests

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -1,21 +1,6 @@
 steps:
   - command:
       - "env"
-      # https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
Sounds sensible to me!

kegsay:
> - Build
> - Unit Tests
> -- wait --
> - Lint
> - SyTests

so build/unit tests are in parallel, along with lint/sytests?

PR #43, opened 2020-02-25
Add a towncrier step in the Sydent CI

babolivier:
Paired with https://github.com/matrix-org/sydent/pull/284

### DIFF
--- a/sydent/pipeline.yml
+++ b/sydent/pipeline.yml
@@ -3,7 +3,7 @@ steps:
       - "echo '--- Install towncrier'"
       - "pip install towncrier"
       - "echo '+++ Check newsfile"
-      - "python -m towncrier.check --compare-with=origin/master"
+      - "python -m towncrier.check"
     label: ":rolled_up_newspaper: Newsfile"
     plugins:
       - docker#v3.0.1:
--- a/sydent/pipeline.yml
+++ b/sydent/pipeline.yml
@@ -1,5 +1,18 @@
 steps:
   - command:
+      - "echo '--- Install towncrier'"
+      - "pip install towncrier"
+      - "echo '+++ Check newsfile"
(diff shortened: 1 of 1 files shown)


sydent/pipeline.yml
```diff
@@ -1,4 +1,17 @@
 steps:
+  - command:
+      - "echo '--- Install towncrier'"
+      - "pip install towncrier"
+      - "echo '+++ Check newsfile"
+      - "python -m towncrier.check --compare-with=origin/master"
```
richvdh: `origin/master` is the default fwiw

sydent/pipeline.yml
```diff
@@ -1,4 +1,17 @@
...
+    label: ":rolled_up_newspaper: Newsfile"
+    plugins:
+      - docker#v3.0.1:
+          image: "python:3.6"
+    branches: "!master !release-*"
+
+  - wait
```
richvdh: I'm a bit unconvinced by this. It's a pain if you have to add a changelog just to check the tests pass on your dev branch.

WDYT?

babolivier: Isn't that the same with Synapse though? The idea here is to avoid using 6 agents when we can know quickly that the CI will need to be re-run for the branch anyway (at least in PRs).

richvdh: it is the same for synapse, but it annoys me there too.

if the CI does get re-run quickly for the branch, the existing build will get cancelled and the agents freed up pretty quickly.

:man_shrugging: Happy to merge this if you like, just wondering if it was an opportunity not to propagate our mistakes.

babolivier: Fair enough, I'll remove it here then

richvdh (approved):

PR #45, opened 2020-02-27
Add pipelines for riot-builder and riot-desktop repos

dbkr:
* riot-desktop already has a buildkite job that can be switched over to use this pipeline and have contrib CI enabled
 * riot-builder is a new repo that needs a buildkite pipeline adding


richvdh (approved):
if you say so

PR #46, opened 2020-02-28
Make dendrite run under sqlite

kegsay:
Should probably work... maybe.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -67,3 +67,33 @@ steps:
           limit: 2
         - exit_status: 2
           limit: 2
+
+  - label: "SyTest - :go: 1.13 / :sqlite: 3"
+    agents:
+      queue: "medium"
+    command:
+      - "bash build.sh"
+      - "bash /bootstrap.sh dendrite"
+    plugins:
+      - docker#v3.0.1:
+          image: "matrixdotorg/sytest-dendrite"
+          propagate-environment: true
+          always-pull: true
+          workdir: "/src"
+          entrypoint: '/bin/sh'
+          shell: ["-x", "-c"]
+          init: false
+          debug: true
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
lgtm from a functional perspective. now we just need the security review

dendrite/pipeline.yml
```diff
@@ -67,3 +67,33 @@ steps:
...
+      - "bash /bootstrap.sh dendrite"
+    plugins:
+      - docker#v3.0.1:
+          image: "matrixdotorg/sytest-dendrite"
+          propagate-environment: true
+          always-pull: true
+          workdir: "/src"
+          entrypoint: '/bin/sh'
+          shell: ["-x", "-c"]
+          init: false
```
richvdh, line 85: I don't know what kind of BS this is

anoadragon453: https://github.com/buildkite-plugins/docker-buildkite-plugin#init-optional-boolean

> Whether or not to run an init process inside the container. This ensures that responsibilities like reaping zombie processes are performed inside the container.

richvdh: Sorry, this wasn't meant to be a particular criticism of this PR - it was a throwaway comment about the state of the pipeline config in general!

But fwiw, I was referring to all three lines here. It seems like an odd way of going about things. The documentation doesn't really make it clear how these things interact with the default entrypoint. Why do we override the default entrypoint at all? Why do we disable init? And if there are good reasons for the above, why are they not written in comments?

I was going to say that we don't have any of this in the synapse pipeline config, but it turns out we do, and it's not any clearer there :/.

Anyway, it certainly doesn't need to block this PR. Just stuff that I think would be good to understand and/or clean up before we cargocult it much further.

PR #52, opened 2020-03-23
Use riot's build-stats and store the output json as artifact

t3chguy:
Requires https://github.com/vector-im/riot-web/pull/12832


richvdh (approved):
:man_shrugging:

PR #53, opened 2020-03-25
Deploy a postgresql alongside sygnal tests

Half-Shot:
In order to run CI against https://github.com/matrix-org/sygnal/pull/91


richvdh (approved):
lgtm but we should presumably wait for support to land in sygnal.

PR #57, opened 2020-04-09
Update dendrite pipeline to lint using xlarge queue

neilalexander:
The linter is a *beast* and requires more RAM to stop it from falling over.


richvdh (approved):
urgh. srsly?

PR #58, opened 2020-04-11
Upgrade instance for android test

BillCarsonFr:
Increased instance size for android test

### DIFF
--- a/riotx-android/pipeline.yml
+++ b/riotx-android/pipeline.yml
@@ -19,9 +19,10 @@ steps:
 
   - label: "Compile Android tests"
     agents:
-      # We use a medium sized instance instead of the normal small ones because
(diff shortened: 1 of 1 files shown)


riotx-android/pipeline.yml
```diff
@@ -19,9 +19,10 @@ steps:
 
   - label: "Compile Android tests"
     agents:
-      # We use a medium sized instance instead of the normal small ones because
+      # We use a xlarge sized instance instead of the normal small ones because
       # gradle build can be memory hungry
-      queue: "medium"
+      # build on the burn-after-use ephemeral queue
```
richvdh: Why? Unless we execute any of the built artifacts, we should use the regular queue,to make more efficient use of the VM

richvdh (approved):

PR #61, opened 2020-04-22
synapse: add step to run sytest with redis

erikjohnston:
I'm not sure this is the best way of going about it, but I've basically munged together the existing docker-compose stuff (used for trial) and the existing sytest stuff.

Is there a way of testing this?

### DIFF
--- a/synapse/docker-compose.sytest.py37.redis.yaml
+++ b/synapse/docker-compose.sytest.py37.redis.yaml
@@ -15,16 +15,8 @@ services:
       WORKERS: "1"
       BLACKLIST: "synapse-blacklist-with-workers"
       REDIS: "redis"
-    always-pull: true
-    workdir: "/src"
-    entrypoint: "/bin/sh"
-    init: false
-    shell: ["-x", "-c"]
-    mount-buildkite-agent: false
+    working_dir: "/src"
+    entrypoint: ""
     volumes:
       - ${BUILDKITE_BUILD_CHECKOUT_PATH}:/src
       - "./logs:/logs"
-    command:
-      - "bash .buildkite/merge_base_branch.sh"
-      - "bash -c 'cat /src/sytest-blacklist /src/.buildkite/worker-blacklist > /src/synapse-blacklist-with-workers'"
-      - "bash /bootstrap.sh synapse"
--- a/synapse/docker-compose.sytest.py37.redis.yaml
+++ b/synapse/docker-compose.sytest.py37.redis.yaml
@@ -1,0 +1,30 @@
+version: '3.1'
+
+services:
+
+  redis:
+    image: redis:5.0
+
+  sytest:
+    image: matrixdotorg/sytest-synapse:py37
+    depends_on:
+      - redis
+    env_file: docker-compose-env
+    environment:
+      POSTGRES: "1"
+      WORKERS: "1"
+      BLACKLIST: "synapse-blacklist-with-workers"
+      REDIS: "redis"
+    always-pull: true
+    workdir: "/src"
+    entrypoint: "/bin/sh"
+    init: false
+    shell: ["-x", "-c"]
+    mount-buildkite-agent: false
+    volumes:
+      - ${BUILDKITE_BUILD_CHECKOUT_PATH}:/src
+      - "./logs:/logs"
+    command:
+      - "bash .buildkite/merge_base_branch.sh"
+      - "bash -c 'cat /src/sytest-blacklist /src/.buildkite/worker-blacklist > /src/synapse-blacklist-with-workers'"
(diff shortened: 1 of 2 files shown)


richvdh:
> Is there a way of testing this?

well, we can make a temporary pipeline which pulls the config from your branch rather than master, but given we'd have to pull the docker-compose config from there too, it's not obvious it's worthwhile.

synapse/docker-compose.sytest.py37.redis.yaml
```diff
@@ -0,0 +1,30 @@
...
+    env_file: docker-compose-env
+    environment:
+      POSTGRES: "1"
+      WORKERS: "1"
+      BLACKLIST: "synapse-blacklist-with-workers"
+      REDIS: "redis"
+    always-pull: true
+    workdir: "/src"
+    entrypoint: "/bin/sh"
+    init: false
```
richvdh: this claims to only be supported as of docker-compose version 3.7: https://docs.docker.com/compose/compose-file/#init

... but I think false is the default, so you can probably just omit it

synapse/docker-compose.sytest.py37.redis.yaml
```diff
@@ -0,0 +1,30 @@
...
+    depends_on:
+      - redis
+    env_file: docker-compose-env
+    environment:
+      POSTGRES: "1"
+      WORKERS: "1"
+      BLACKLIST: "synapse-blacklist-with-workers"
+      REDIS: "redis"
+    always-pull: true
+    workdir: "/src"
```
richvdh: these aren't valid compose-file options.

synapse/docker-compose.sytest.py37.redis.yaml
```diff
@@ -0,0 +1,30 @@
...
+    entrypoint: "/bin/sh"
+    init: false
+    shell: ["-x", "-c"]
+    mount-buildkite-agent: false
```
richvdh: these aren't valid options either.

richvdh (changes requested):

richvdh:
> it's not obvious it's worthwhile.

scratch that: there's enough novel stuff here that we should try it out on a branch first. If you can update the url for the docker-compose file to point to your branch, I'll set up a test pipeline.

Before you do that though, it's worth testing the docker-compose file locally. Currently:

```
rav@fred:~/work/pipelines/synapse (erikj/redis_test_synapse $=)$ BUILDKITE_BUILD_CHECKOUT_PATH=$HOME/work/synapse docker-compose -f docker-compose.sytest.py37.redis.yaml run sytest
ERROR: The Compose file './docker-compose.sytest.py37.redis.yaml' is invalid because:
Unsupported config option for services.sytest: 'mount-buildkite-agent'
```

synapse/docker-compose.sytest.py37.redis.yaml
richvdh: (you don't need this with the docker-compose buildkite plugin, because it doesn't suffer the same foot-gun as the docker buildkite plugin where failing to set it leaks the buildkite agent token into the container. See also https://github.com/matrix-org/internal-config/wiki/BuildKite#not-running-untrusted-code-on-the-build-agents.)

erikjohnston:
> > it's not obvious it's worthwhile.
> 
> scratch that: there's enough novel stuff here that we should try it out on a branch first. If you can update the url for the docker-compose file to point to your branch, I'll set up a test pipeline.

Done!

> Before you do that though, it's worth testing the docker-compose file locally. Currently:
> 
> ```
> rav@fred:~/work/pipelines/synapse (erikj/redis_test_synapse $=)$ BUILDKITE_BUILD_CHECKOUT_PATH=$HOME/work/synapse docker-compose -f docker-compose.sytest.py37.redis.yaml run sytest
> ERROR: The Compose file './docker-compose.sytest.py37.redis.yaml' is invalid because:
> Unsupported config option for services.sytest: 'mount-buildkite-agent'
> ```

Aha, thanks!

erikjohnston:
Gah, i shouldn't have review requested again while testing, sowwy. But looks to be ready now.

Small thing I don't really understand, I had to mount `/tmp/logs` rather than `./logs` as no logs where found during the upload artefact step in the latter case.

richvdh:
> Small thing I don't really understand, I had to mount /tmp/logs rather than ./logs as no logs where found during the upload artefact step in the latter case.

I think the basic problem is that docker-compose doesn't let you use relative paths for volumes. The other buildkite steps work because the docker-buildkite-plugin makes it absolute for you (https://github.com/buildkite-plugins/docker-buildkite-plugin#volumes-optional-array-or-boolean).

richvdh (approved):

PR #66, opened 2020-05-19
Fix typo in towncrier's script for Sydent

babolivier:
Not entirely sure why it's only coming up now though, as it makes Sydent's CI fail (see https://buildkite.com/matrix-dot-org/sydent/builds/492#1dda068b-4424-45fe-b529-63af4ad82cc1).


clokep (approved):
This looks fine, not sure if I can actually approve it... 🤷

PR #68, opened 2020-05-22
Dendrite tests with full HTTP APIs

neilalexander:
The Dendrite monolith mode normally short-circuits the internal APIs and calls the functions directly, but this means that the HTTP versions of these APIs as used by the polylith components are never tested under Sytest. 

This adds two new stages to the pipeline to use the new full API mode for the Dendrite monolith which will use the HTTP APIs rather than short-circuiting, as this will give us testing that is quite a good emulation of running separate polylith components without actually having to.

They are currently soft-fail. I will update them to fail the pipeline hard once some Dendrite issues are fixed.

Related: matrix-org/dendrite#1057, matrix-org/sytest#877.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -68,7 +68,7 @@ steps:
         - exit_status: 2
           limit: 2
   
-   - label: "SyTest - :go: 1.13 / :postgres: 9.6 / full HTTP APIs"
+  - label: "SyTest - :go: 1.13 / :postgres: 9.6 / full HTTP APIs"
     agents:
       queue: "medium"
     env:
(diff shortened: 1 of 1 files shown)


kegsay (approved):
I worry a little about doing lots of work on each commit, as it takes long enough as it is, but this is more of an infrastructure issue...

dendrite/pipeline.yml
```diff
@@ -67,6 +67,40 @@ steps:
           limit: 2
         - exit_status: 2
           limit: 2
+  
+  - label: "SyTest - :go: 1.13 / :postgres: 9.6 / full HTTP APIs"
```
Half-Shot, line 71: Sad noises about 9.6, is *anyone* actually running a 9.6 with dendrite these days?

richvdh (approved):
seems legit

PR #70, opened 2020-06-15
Don't soft-fail postgres Dendrite sytests

neilalexander:
This removes `soft_fail` on the auxiliary Dendrite tests now as they should be more reliable.


kegsay:
Just do postgres for now please, as there are known flakey tests on sqlite https://github.com/matrix-org/dendrite/issues/1099

neilalexander:
Updated.

kegsay (approved):

richvdh (approved):

PR #74, opened 2020-07-09
Prevent mounting BK agent for Sydent builds

anoadragon453:
This is done in an effort to allow community PRs to run in CI for Sydent. I compared the pipeline with [synapse's](https://github.com/matrix-org/pipelines/blob/master/synapse/pipeline.yml).

I'm currently unsure whether `propagate-environment` is necessary for Sydent, my guess is that it's necessary if the CI is running any scripts, which Sydent's CI currently does not do.


richvdh (approved):
lgtm

PR #81, opened 2020-07-24
Synapse: Add step to check for unix file endings

anoadragon453:
To address https://github.com/matrix-org/synapse/issues/7943. I'm using the `python:3.6` image for consistency, but this would probably run on `alpine`.

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -55,7 +55,7 @@ steps:
       # Run 'file' command on each file, checking for Windows-style line terminators.
       # grep returns exit code 1 when it doesn't find anything. We invert this as
       # finding entries are a bad thing
-      - "find . -type f -exec file '{}' ';' | (! grep 'CRLF line terminators')"
+      - "find . -type f -path './.git/*' -prune -o -exec file '{}' ';' | (! grep 'CRLF line terminators')"
     plugins:
       - docker#v3.0.1:
           image: "python:3.6"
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -50,6 +50,17 @@ steps:
           image: "python:3.6"
           mount-buildkite-agent: false
 
+  - label: "\U0001F5A5 check unix line-endings"
+    command:
+      # Run 'file' command on each file, checking for Windows-style line terminators.
+      # grep returns exit code 1 when it doesn't find anything. We invert this as
+      # finding entries are a bad thing
+      - "find . -type f -path './.git/*' -prune -o -exec file '{}' ';' | (! grep 'CRLF line terminators')"
```
richvdh: the trouble with this is that I'm not sure what makes `file` decide to print "CRLF line terminators", and I'd kinda like it to complain if there is a CRLF terminator anywhere in the file, even if they aren't used consistently throughout. I'd just use grep:

```suggestion
      - "find . -path './.git/*' -prune -o -type f -print0 | xargs -0 grep -l $'\r$' && { echo 'found files with CRLF line endings'; exit 1 }"
```

It might also be worth sticking that in a script in the synapse repo rather than here.

anoadragon453: https://github.com/matrix-org/synapse/pull/7965

anoadragon453:
Updated to use the new repo script :slightly_smiling_face:

richvdh (approved):
lgtm.

part of me wonders if we should merge some of these lint steps to save spinning up lots of buildkite agents... otoh we're going to need lots of buildkite agents to run the tests so it probably doesn't make much difference.

PR #82, opened 2020-07-30
Synapse: Link to the contributing guide if newsfragment check fails

anoadragon453:
There are now multiple cases of community contributors being confused about what to do when the newsfragment check fails:

* https://github.com/matrix-org/synapse/pull/7906#issuecomment-661086316
* https://github.com/matrix-org/synapse/pull/7314#issuecomment-666253337

When `python -m towncrier.check --compare-with=origin/develop` fails, it exits with error code `1`, otherwise exit code `0` when the check succeeds. We use this to `echo` a link to the changelog contributing guide when the check fails, to hopefully allow contributors to be informed and solve the issue themselves.

This was originally a [Synapse PR](https://github.com/matrix-org/synapse/pull/8002), until I realized that the pipeline calls a bash script instead, which does its own checks. Thus we'd want to print the guide on those failing as well.

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -33,7 +33,7 @@ steps:
   - label: ":newspaper: Newsfile"
     command:
       - "python -m pip install tox"
-      - "scripts-dev/check-newsfragment"
+      - "scripts-dev/check-newsfragment || echo 'Please see the contributing guide for help: https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog'"
     branches: "!master !develop !release-*"
     plugins:
       - docker#v3.0.1:


synapse/pipeline.yml
```diff
@@ -33,7 +33,7 @@ steps:
   - label: ":newspaper: Newsfile"
     command:
       - "python -m pip install tox"
-      - "scripts-dev/check-newsfragment"
+      - "scripts-dev/check-newsfragment || echo 'Please see the contributing guide for help: https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog'"
```
richvdh, line 36: I suspect this will exit with a zero exitcode on failure, so the buildstep won't reject invalid newsfragments.

tbh I think you were probably right the first time: it belongs in the script, not here.

richvdh (changes requested):

synapse/pipeline.yml
anoadragon453: [The script](https://github.com/matrix-org/synapse/blob/develop/scripts-dev/check-newsfragment) has `set -e`, so any non-zero exit on a command will exit the script with a nonzero exit code. And the script specifically specifies `exit 1` for any errors it encounters.

Putting it in the script would require some refactoring of the script to cover all error cases.

richvdh: > Putting it in the script would require some refactoring of the script to cover all error cases.

Maybe? It's not actually obvious to me we want to print this out every time the script exits non-zero.

The script already has user-friendly error messages for the cases where it's the user's fault rather than something else. I suggest it makes more sense to improve those errors than it does to add *more* text here. 

If the goal is to avoid repeating the "Please see the guide..." text, just stick that in a variable in the script so that it can be referenced in each error message.

anoadragon453: I see what you mean. We indeed wouldn't want `Please see the contributing guide...` to appear if `git fetch -q origin develop` fails.

I think that a link to the contributing guide being included on the tail end of the each of the errors in the script does make sense though, especially if the user has made multiple errors. They should read the guide and fix all of them instead of going through each via trial-and-error.

I'll try to modify the script instead, including detected `tox`'s error code inside the script so the link text output is limited to that failing, rather than other random bits of the script. Thanks for bearing with me on this :slightly_smiling_face:

anoadragon453:
Superseded by https://github.com/matrix-org/synapse/pull/8024

PR #83, opened 2020-08-04
Add pipelines for iOS projects

AliSoftware:
Add 3 new BuildKit pipelines, for the eponym repositories:

 - `matrix-ios-sdk`
 - `matrix-ios-kit`
 - `element-ios`

⚠️ Once this PR is merged, we need to edit the 3 BuildKite pipelines' configurations to point to those `yml` files in `master` (as BK is set up to currently point to `oha/matrix-ios-sdk` branch)

_Note: In the future, we plan to have a Scheduled Build configured on `matrix-ios-sdk`, which would trigger every night (at 2am? 3am?), with the `NIGHTLY_BUILD=1` env var set. That is what the `&nightly` condition is there for_


element-ios/pipeline.yml
```diff
@@ -0,0 +1,58 @@
...
+        build.branch !~ /^release\// &&
+        build.env("NIGHTLY_BUILD") == null
+    - &nightly |
+      // Only for nightly builds
+      build.env("NIGHTLY_BUILD") != null
+  commands:
+    - &setup |
+      brew install cmake
+      bundle install
+      printf %100s\\n | tr ' ' '='
```
richvdh: worth noting that buildkite has facilities for splitting up log output into sections, which may be superior here. See https://buildkite.com/docs/pipelines/managing-log-output

element-ios/pipeline.yml
```diff
@@ -0,0 +1,58 @@
+# This is just a dummy entry (the `x-yaml-aliases` key is not an official pipeline key, and will be ignored by BuildKite)
+#   that we use only to store YAML anchors (`&xxx`), that we plan to use and reference later in the YAML file (using `*xxx`)
+#   without having to copy/paste the same values over and over.
+#   Note: keys like `agent`, `env`, … used here are totally arbitrary; the only point is to define various separate `&xxx` anchors there.
+#
```
richvdh: think you have some spurious indentation here (and in the other files)

```suggestion
# that we use only to store YAML anchors (`&xxx`), that we plan to use and reference later in the YAML file (using `*xxx`)
# without having to copy/paste the same values over and over.
# Note: keys like `agent`, `env`, … used here are totally arbitrary; the only point is to define various separate `&xxx` anchors there.
#
```

element-ios/pipeline.yml
```diff
@@ -0,0 +1,58 @@
...
+    - &nightly |
+      // Only for nightly builds
+      build.env("NIGHTLY_BUILD") != null
+  commands:
+    - &setup |
+      brew install cmake
+      bundle install
+      printf %100s\\n | tr ' ' '='
+    - &build_app
+      bundle exec fastlane build
```
richvdh: I'm not convinced that using yaml anchors quite so widely is good for either readability or maintainability (imho you may as well write `bundle exec fastlane build` as `*build_app`, and only switch to yaml anchors when it gets more complicated), but I guess it's fine.

element-ios/pipeline.yml
AliSoftware: Nice! didn't know about that, I'll make the change 👍

element-ios/pipeline.yml
AliSoftware: yeah I might get overexcited with anchors ^^ it was quite useful when debugging the pipeline as it's then easy to change the command just here (to an `echo` for debug, or to run another fastlane lane, etc) in a single place and have it applied in both the steps for Feature branch and release branch.

That's also in preparation for future changes in these pipelines, as for now we do only a few steps (because the single MacMini CI machine that we have is not super powerful so we don't want to put it to its knees) but in the future we plan to add other steps, like building the same commands on master and develop branches too etc, so that will get repeated again when we introduce those.

element-ios/pipeline.yml
AliSoftware: Fixed, it's indeed much better, thanks for the tip!

element-ios/pipeline.yml
AliSoftware: Turns out it was useful to factorize the `echo --- some section title` in each cases too after all 😜

richvdh (dismissed):
lgtm

matrix-ios-kit/pipeline.yml
```diff
@@ -0,0 +1,88 @@
...
+        build.env("NIGHTLY_BUILD") == null
+    - &nightly |
+      // Only for nightly builds
+      build.env("NIGHTLY_BUILD") != null
+  commands:
+    - &setup |
+      echo --- Setup
+      brew install cmake
+      bundle install
+      printf %100s\\n | tr ' ' '='
```
AliSoftware: Whoops forgot to remove that one

AliSoftware:
```suggestion
```

richvdh (approved):

PR #84, opened 2020-08-05
Remove build.sh from Dendrite sytest pipelines

neilalexander:
Our `dendrite_sytest.sh` bootstrap script already builds only exactly what Sytest needs to run - the monolith server and `generate-keys` - so we don't need to run `build.sh` here too. It just ends up taking a long time to build all of the components that don't get used during Sytest.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -1,7 +1,7 @@
 steps:
   - command:
       - "env"
-      - "go build ./cmd/..."
+      - "bash build.sh"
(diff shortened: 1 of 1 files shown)


kegsay:
It's valuable to still run build.sh as it checks other components compile correctly e.g dendritejs, polylith components.

neilalexander:
If we want to continue running `build.sh` then we should really do that in the initial build step. We don't need to repeat it four times for sytest :D

kegsay:
Fair point!

kegsay (approved):

PR #86, opened 2020-08-06
Use the redis server in the sytest image

richvdh:
As of https://github.com/matrix-org/sytest/pull/928, the sytest-synapse images
support a redis server, so we no longer need to docker-compose.


erikjohnston (approved):
I think this works. After merge we should double check the test output logs to make sure it is connecting to redis

PR #87, opened 2020-08-21
Run complement as part of Dendrite pipeline

kegsay:
Tested manually with `bk local run just-this-command.yml` and it all works as intended. There's a number of outstanding questions:
 - Security: This requires the docker socket to be forwarded through to Complement. Can we make some guarantees that this is safe (e.g spot instances?)
 - Security: This uses a [convenience script](https://docs.docker.com/engine/install/debian/#install-using-the-convenience-script) to install Docker, is this okay?
 - Concurrency: When a homeserver tries to talk to Complement it tries to hit `:8448`, which is now published so this can happen. This is a problem if we want to run multiple Complement instances on a single machine. See under using spot instances.
 - ~~Security: Currently this pipeline will execute `build/scripts/complement.sh` which is a script inside of the Dendrite repository. I want to run Complement on Dendrite community PRs - does the fact that we pull the script from Dendrite cause an issue? If so, should I just inline the script in this pipeline? It's 6 lines long.~~

### DIFF
--- a/dendrite/Complement.Dockerfile
+++ b/dendrite/Complement.Dockerfile
@@ -1,0 +1,7 @@
+# This Dockerfile prepares an image with Complement/Docker pre-installed.
+# This allows users of this image to issue `docker build` commands to build their HS
+# and then run Complement against it.
+FROM golang:1.15-buster
+RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
+ADD https://github.com/matrix-org/complement/archive/master.tar.gz .
+RUN tar -xzf master.tar.gz && cd complement-master && go mod download
--- a/dendrite/Complement.Dockerfile
+++ b/dendrite/Complement.Dockerfile
@@ -1,7 +1,0 @@
-# This Dockerfile prepares an image with Complement/Docker pre-installed.
-# This allows users of this image to issue `docker build` commands to build their HS
-# and then run Complement against it.
-FROM golang:1.15-buster
-RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
-ADD https://github.com/matrix-org/complement/archive/master.tar.gz .
-RUN tar -xzf master.tar.gz && cd complement-master && go mod download
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -44,6 +44,7 @@ steps:
       queue: "medium"
     plugins:
       - docker#v3.5.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
           image: "matrixdotorg/complement:latest"
           mount-buildkite-agent: false
           # Complement needs to know if it is running under CI
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -35,6 +35,22 @@ steps:
       - docker#v3.0.1:
           image: "golang:1.13"
           mount-buildkite-agent: false
+
+  - command:
+      - "curl -fsSL https://get.docker.com -o get-docker.sh"
+      - "sh get-docker.sh"
+      - "bash build/scripts/complement.sh"
+    label: "\U0001F9EA Complement"
+    agents:
+      queue: "medium"
+    plugins:
+      - docker#v3.5.0:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
+          propagate-environment: true
+          publish: [ "8448:8448" ]
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
           
   - label: "SyTest - :go: 1.13 / :postgres: 9.6"
     agents:
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -44,7 +44,7 @@ steps:
       queue: "medium"
     plugins:
       - docker#v3.5.0:
-          image: "kegsay/complement:latest"
+          image: "matrixdotorg/complement:latest"
           mount-buildkite-agent: false
           # Complement needs to know if it is running under CI
           environment:
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -49,8 +49,12 @@ steps:
       - docker#v3.5.0:
           image: "golang:1.13"
           mount-buildkite-agent: false
-          propagate-environment: true
+          # Complement needs to know if it is running under CI
+          environment:
+           - "CI=true"
           publish: [ "8448:8448" ]
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the hosts Docker.
           volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
           
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -39,7 +39,9 @@ steps:
   - command:
       - "curl -fsSL https://get.docker.com -o get-docker.sh"
       - "sh get-docker.sh"
-      - "bash build/scripts/complement.sh"
+      - "docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile ."
+      - "wget https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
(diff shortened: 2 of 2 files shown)


anoadragon453 (commented):
> Security: Currently this pipeline will execute `build/scripts/complement.sh` which is a script inside of the Dendrite repository. I want to run Complement on Dendrite community PRs - does the fact that we pull the script from Dendrite cause an issue? If so, should I just inline the script in this pipeline? It's 6 lines long.

I don't believe so as it'll be executed inside of the docker container alongside any other untrusted code from a PR.

If it is a problem, then we have a few other pipelines to change :)

kegsay:
> I don't believe so as it'll be executed inside of the docker container alongside any other untrusted code from a PR.
>
> If it is a problem, then we have a few other pipelines to change :)

In this particular case it is an issue potentially because we are mounting the docker socket from the host - most pipelines do not do this. This means untrusted code can issue `docker` queries, and in the worst case run a container which they control (but uploaded to docker hub) which also mounts the entire host `/` which could be used to exfiltrate passwords, etc.

anoadragon453:
> In this particular case it is an issue potentially because we are mounting the docker socket from the host - most pipelines do not do this. This means untrusted code can issue docker queries, and in the worst case run a container which they control (but uploaded to docker hub) which also mounts the entire host / which could be used to exfiltrate passwords, etc.

Ah, well I guess the main question there is if we have any secrets/passwords in the host box that we're worried about?

anoadragon453:
> Ah, well I guess the main question there is if we have any secrets/passwords in the host box that we're worried about?

Looks like our buildkite agent token is stored in these containers, so yes we wouldn't want contributor coded to be able to run in them.

dendrite/pipeline.yml
```diff
@@ -35,6 +35,22 @@ steps:
       - docker#v3.0.1:
           image: "golang:1.13"
           mount-buildkite-agent: false
+
+  - command:
+      - "curl -fsSL https://get.docker.com -o get-docker.sh"
+      - "sh get-docker.sh"
+      - "bash build/scripts/complement.sh"
```
neilalexander: I agree that we don't want to pull this script from the repository - for the small number of lines, let's inline this.

dendrite/pipeline.yml
```diff
@@ -35,6 +35,24 @@ steps:
       - docker#v3.0.1:
           image: "golang:1.13"
           mount-buildkite-agent: false
+
+  - command:
+      - "curl -fsSL https://get.docker.com -o get-docker.sh"
+      - "sh get-docker.sh"
+      - "docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile ."
+      - "wget https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
```
richvdh: would it be possible to build a complement docker image, which includes both the docker commandline and a built (or at least ready-to-build) complement, so we don't need to do this messing about?

dendrite/pipeline.yml
```diff
@@ -35,6 +35,24 @@ steps:
...
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v ./tests"
+    label: "\U0001F9EA Complement"
+    agents:
+      queue: "medium"
+    plugins:
+      - docker#v3.5.0:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
+          propagate-environment: true
```
richvdh: why do we need this? (other than for cargo-culting from other steps which also don't need it?)

dendrite/pipeline.yml
```diff
@@ -35,6 +35,24 @@ steps:
...
+          publish: [ "8448:8448" ]
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
```
richvdh: it'd be god to add comments explaining why we need these

richvdh (changes requested):

dendrite/pipeline.yml
kegsay: Yes that's possible.

dendrite/pipeline.yml
kegsay: Because Complement relies on `CI=true` to determine if it should be using `localhost` or the docker `172` IP when communicating with homeservers.

richvdh: ok, a comment explaining that would be good.

kegsay: To be fair I can just manually do this, so will remove this line.

dendrite/pipeline.yml
kegsay: This is actually a bit of a faff to do. I want Complement to be built from `master` every time, and I want that build to include the Docker CLI so I can issue `docker build` on Dendrite.

We're debating where to put docker images full stop so for now I think I'll leave this and we can optimise it later.

richvdh: docker images are still going on dockerhub for now.

Can you at least give us an image so that we don't need to do the awful get-docker thing?

kegsay: Sure, I'll upload one with Go/Docker CLI installed already.

kegsay: Done, which shaves 1min off the runtime.

dendrite/Complement.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# This Dockerfile prepares an image with Complement/Docker pre-installed.
+# This allows users of this image to issue `docker build` commands to build their HS
+# and then run Complement against it.
+FROM golang:1.15-buster
+RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
+ADD https://github.com/matrix-org/complement/archive/master.tar.gz .
+RUN tar -xzf master.tar.gz && cd complement-master && go mod download
```
richvdh: I'm confused. why do we need to download complement here as well as in the build steps?

(and if we're building a complement-specific image, it should probably live in the complement repo, not hidden here?)

dendrite/pipeline.yml
```diff
@@ -35,6 +35,25 @@ steps:
...
+
+  - command:
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v ./tests"
+    label: "\U0001F9EA Complement"
+    agents:
+      queue: "medium"
+    plugins:
+      - docker#v3.5.0:
+          image: "kegsay/complement:latest"
```
richvdh: please can this live in the matrixdotorg namespace.

richvdh (changes requested):

dendrite/Complement.Dockerfile
kegsay: We download Complement here to install dependencies to slightly speed up builds (hence why we just do `go mod download` and don't run Complement itself), so deps are baked into the image. I'm not comfortable making a complement-specific image at this point as I don't know what other requirements may sprout yet, so I'd rather hide it here for this specific instance.

richvdh: I don't really follow. It's *already* a complement-specific image. It may not work fully or be ready for use by others, but I don't see why those problems are solved by keeping the dockerfile here?

kegsay: Because if people see a Dockerfile in the complement repo people think that:
 - it works fully.
 - it's ready to use by others.

when neither are true. I can add appropriate warnings but it seems easier to leave it in the place which required the creation of the Dockerfile in the first place.

dendrite/pipeline.yml
```diff
@@ -35,6 +35,25 @@ steps:
...
+
+  - command:
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v ./tests"
+    label: "\U0001F9EA Complement"
+    agents:
+      queue: "medium"
+    plugins:
+      - docker#v3.5.0:
+          image: "matrixdotorg/complement:latest"
```
richvdh, line 48:
```suggestion
          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
          image: "matrixdotorg/complement:latest"
```

richvdh (approved):
lgtm!

PR #92, opened 2020-08-28
Build dendrite for complement

kegsay:
Whoops, forgot to actually build the thing.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -37,6 +37,7 @@ steps:
           mount-buildkite-agent: false
 
   - command:
+      - docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile .
       - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
       - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v ./tests"
     label: "\U0001F9EA Complement"


neilalexander (approved):

dendrite/pipeline.yml
```diff
@@ -37,6 +37,7 @@ steps:
           mount-buildkite-agent: false
 
   - command:
+      - docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile .
```
richvdh, line 40: Camelcase, ugh. :man_shrugging:

dendrite/pipeline.yml
```diff
@@ -37,6 +37,7 @@ steps:
...
       - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
       - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v ./tests"
```
richvdh, line 42: I would say that at the point you have five commands to run, you should consider putting a script in the `dendrite` repo and using that instead.

but this is fine for now.

richvdh (approved):

dendrite/pipeline.yml
kegsay: It was originally a script in the Dendrite repo but we decided against it because it would then be under the control of anyone in a PR.

PR #97, opened 2020-09-15
Add soft-failing Complement to the Synapse pipeline

anoadragon453:
This PR adds a Complement build step to the Synapse pipeline. It is currently marked as "soft-fail" (meaning this step failing will not block the whole build) as not all complement tests work with Synapse yet.

It is modeled after the existing [Dendrite pipeline entry](https://github.com/matrix-org/pipelines/blob/aa6694b6cde1b5ef5464b6683b8dee5c685ab2a8/dendrite/pipeline.yml#L39-L58), but instead moves the commands to [a script](https://github.com/matrix-org/synapse/blob/ed858b1f97f49f928b4a6e8de7d040bb6f44210b/scripts-dev/complement.sh) inside the Synapse repo.

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -535,9 +535,14 @@ steps:
   ################################################################################
 
   - command:
+      # Build a docker image from the checked out Synapse source
       - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
+      # Download Complement's source code
       - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      # Build a second docker image on top of the above image. This one sets up Synapse with a generated config file,
+      # signing and SSL keys so Synapse can run and federate
       - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
+      # Finally, run the tests. Complement has already been built in the matrixdotorg/complement:latest" image which we're running in
       - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
     label: "\U0001F9EA Complement"
     agents:
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -537,13 +537,16 @@ steps:
   - command:
       # Build a docker image from the checked out Synapse source
       - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
-      # Download Complement's source code
-      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      # We use the complement:latest image to provide Complement's dependencies, but want
+      # to actually run against the latest version of Complement, so download it here.
+      - "wget https://github.com/matrix-org/complement/archive/master.tar.gz"
+      - "tar -xzf master.tar.gz"
       # Build a second docker image on top of the above image. This one sets up Synapse with a generated config file,
       # signing and SSL keys so Synapse can run and federate
       - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
-      # Finally, run the tests. Complement has already been built in the matrixdotorg/complement:latest" image which we're running in
-      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
+      # Finally, compile and run the tests.
+      - "cd complement-master"
+      - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
     label: "\U0001F9EA Complement"
     agents:
       queue: "medium"
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -535,7 +535,10 @@ steps:
   ################################################################################
 
   - command:
-      - "bash scripts-dev/complement.sh"
+      - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
     label: "\U0001F9EA Complement"
     agents:
       queue: "medium"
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -527,3 +527,31 @@ steps:
...
 #    command: "curl -k https://coveralls.io/webhook?repo_token=$COVERALLS_REPO_TOKEN -d \"payload[build_num]=$BUILDKITE_BUILD_NUMBER&payload[status]=done\""
+
+  ################################################################################
+  #
+  # Complement Test Suite
+  #
+  ################################################################################
+
+  - command:
+      - "bash scripts-dev/complement.sh"
```
kegsay: Careful, as this allows nasty Synapse PRs to execute arbitrary stuff.

anoadragon453: Ah, yes classic. I've just move the commands into the pipeline then.

synapse/pipeline.yml
```diff
@@ -527,3 +527,34 @@ steps:
...
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          # Complement needs to know if it is running under CI
+          environment:
+           - "CI=true"
+          publish: [ "8448:8448" ]
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the hosts Docker.
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
```
dbkr, line 568: hmm - I don't really know enough about docker to reason whether this is safe or not. Would it not let you start inspecting other running containers from other buildkite jobs and/or killing them off?

anoadragon453: That's correct, and I'm mostly leaning on the fact that Dendrite does the same thing and a security discussion has already happened [here](https://github.com/matrix-org/pipelines/pull/87) :D

richvdh: yes it would. We're trusting the complement master branch and docker image not to do nasty things.

It looks like a difference between https://github.com/matrix-org/pipelines/pull/87 and here is that we are running a `docker build` for the synapse branch under test within the privileged container. I suspect that could be used to construct a container escape, which is problematic.

I don't know where the dendrite docker imaged used in #87 comes from.

synapse/pipeline.yml
```diff
@@ -527,3 +527,34 @@ steps:
...
+  #
+  # Complement Test Suite
+  #
+  ################################################################################
+
+  - command:
+      - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
```
richvdh: can we have a comment describing what this is doing and why?

For instance: why do we pull a docker image of `complement:latest`, but then download a separate tarball, unpack it, and run that?

why do we build two "synapse" docker images?

anoadragon453: I've added some comments, the process is indeed a bit complicated. I think it could be cleaned up actually, perhaps by mounting the Synapse checkout instead of building one image from another. But not sure whether we want to do that atm.

I'm also not sure how long this will take to run on our build infra. If it's longer than Sytest is currently then I'd probably advocate for putting in the work to mount instead of building a Synapse container from scratch.

synapse/pipeline.yml
anoadragon453: Right, a community member reminded me that this was added in https://github.com/matrix-org/pipelines/pull/92, and is built from the Dendrite repo. Is this a concern?

synapse/pipeline.yml
```diff
@@ -527,3 +527,39 @@ steps:
...
+  ################################################################################
+  #
+  # Complement Test Suite
+  #
+  ################################################################################
+
+  - command:
+      # Build a docker image from the checked out Synapse source
+      - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
+      # Download Complement's source code
```
richvdh: There's an argument that this incantation belongs inside the complement docker image (in the same way that sytest does it here: https://github.com/matrix-org/sytest/blob/develop/docker/bootstrap.sh#L32). But for now let's at least document what it's doing more clearly:

```suggestion
      # We use the complement:latest image to provide Complement's dependencies, but want
      # to actually run against the latest version of Complement, so download it here.
```

synapse/pipeline.yml
```diff
@@ -527,3 +527,39 @@ steps:
...
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
+      # Build a second docker image on top of the above image. This one sets up Synapse with a generated config file,
+      # signing and SSL keys so Synapse can run and federate
+      - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
+      # Finally, run the tests. Complement has already been built in the matrixdotorg/complement:latest" image which we're running in
```
richvdh: well no it hasn't, and even if it had, we're not using the version from the image; we're using the version we downloaded. `go test` does the compilation automatically.

```suggestion
      # Finally, compile and run the tests.
```

synapse/pipeline.yml
```diff
@@ -527,3 +527,39 @@ steps:
...
+      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
```
```suggestion
      - "cd complement-master"
      - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v ./tests"
```

synapse/pipeline.yml
richvdh: it's fine; I'm an idiot. `docker build`, of course, already runs all of its commands in an isolated container.

synapse/pipeline.yml
```diff
@@ -527,3 +527,39 @@ steps:
...
+  #
+  # Complement Test Suite
+  #
+  ################################################################################
+
+  - command:
+      # Build a docker image from the checked out Synapse source
+      - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
+      # Download Complement's source code
+      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
```
richvdh: `-N` is redundant (we don't expect to already have a `master.tar.gz`; indeed its presence causes a warning), and since we're already in a list of commands, we may as well use it instead of our own `&&`.

```suggestion
      - "wget https://github.com/matrix-org/complement/archive/master.tar.gz"
      - "tar -xzf master.tar.gz"
```

richvdh (dismissed):

anoadragon453:
Those all made sense, thanks!

richvdh (approved):

PR #102, opened 2020-10-17
add lint analysis for the sdk source

bmarty:


element-android/pipeline.yml
```diff
@@ -68,6 +68,23 @@ steps:
           propagate-environment: true
           mount-buildkite-agent: false
 
+  - label: ":android: Lint analysis of the SDK"
+    agents:
+      # We use a xlarge sized instance instead of the normal small ones because
+      # gradle build can be memory hungry
+      # build on the burn-after-use ephemeral queue
+      queue: "ephemeral-xlarge"
```
richvdh: unless you are building runnable artifacts, I'd suggest the regular `xlarge` queue which should be a bit faster.

bmarty: If you say so. I've made the change

richvdh (approved):
lgtm

PR #103, opened 2020-10-22
Do not wait for the whole steps to be run, or even successful to start the εxodus analysis

bmarty:
This PR is to avoid the εxodus not to be run if an unnecessary step fails, for instance see https://buildkite.com/matrix-dot-org/element-android/builds/484 :

image

But I'm really not confident to the changes I've made, because I do not know how to test the new pipeline.

Inspired from https://buildkite.com/docs/pipelines/dependencies

### DIFF
--- a/element-android/pipeline.yml
+++ b/element-android/pipeline.yml
@@ -39,7 +39,7 @@ steps:
     commands:
       - "./gradlew clean lintGplayRelease assembleGplayDebug --stacktrace"
       - "mv vector/build/outputs/apk/gplay/debug/*.apk ."
-      key: "elementApp"
+    key: "elementApp"
     artifact_paths:
       - "*.apk"
       - "vector/build/reports/*.*"
@@ -77,7 +77,7 @@ steps:
     commands:
       - "./gradlew clean assembleGplayRelease --stacktrace"
       - "mv vector/build/outputs/apk/gplay/release/*.apk ."
-      key: "elementApp"
+    key: "elementApp"
     artifact_paths:
       - "*.apk"
     branches: "master"
--- a/element-android/pipeline.yml
+++ b/element-android/pipeline.yml
@@ -39,6 +39,7 @@ steps:
     commands:
       - "./gradlew clean lintGplayRelease assembleGplayDebug --stacktrace"
       - "mv vector/build/outputs/apk/gplay/debug/*.apk ."
+      key: "elementApp"
     artifact_paths:
       - "*.apk"
       - "vector/build/reports/*.*"
@@ -76,6 +77,7 @@ steps:
     commands:
       - "./gradlew clean assembleGplayRelease --stacktrace"
       - "mv vector/build/outputs/apk/gplay/release/*.apk ."
+      key: "elementApp"
     artifact_paths:
       - "*.apk"
     branches: "master"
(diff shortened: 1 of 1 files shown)


element-android/pipeline.yml
```diff
@@ -39,6 +39,7 @@ steps:
     commands:
       - "./gradlew clean lintGplayRelease assembleGplayDebug --stacktrace"
       - "mv vector/build/outputs/apk/gplay/debug/*.apk ."
+      key: "elementApp"
```
richvdh: I think this needs to be in the parent object:

```suggestion
    key: "elementApp"
```

the same for the other two places below.

richvdh (changes requested):

element-android/pipeline.yml
```diff
@@ -127,3 +127,4 @@ steps:
       - "mv vector-gplay-universal-*.apk app.apk"
       - "docker run -v $(pwd)/app.apk:/app.apk --rm -i exodusprivacy/exodus-standalone -j > exodus.json"
       - "jq -e '.trackers == []' exodus.json > /dev/null || { echo 'static analysis identified user tracking library' ; false; }"
+      depends_on: "elementApp"
```
richvdh:
```suggestion
    depends_on: "elementApp"
```

element-android/pipeline.yml
```diff
@@ -39,6 +39,7 @@ steps:
     commands:
       - "./gradlew clean lintGplayRelease assembleGplayDebug --stacktrace"
       - "mv vector/build/outputs/apk/gplay/debug/*.apk ."
+    key: "elementApp"
```
richvdh: we should surely not have two steps with the same key?

```suggestion
    key: "gplay_debug"
```

element-android/pipeline.yml
```diff
@@ -92,6 +93,7 @@ steps:
     commands:
       - "./gradlew clean assembleGplayRelease --stacktrace"
       - "mv vector/build/outputs/apk/gplay/release/*.apk ."
+    key: "elementApp"
```
richvdh:
```suggestion
    key: "gplay_release"
```

element-android/pipeline.yml
```diff
@@ -143,3 +143,4 @@ steps:
       - "mv vector-gplay-universal-*.apk app.apk"
       - "docker run -v $(pwd)/app.apk:/app.apk --rm -i exodusprivacy/exodus-standalone -j > exodus.json"
       - "jq -e '.trackers == []' exodus.json > /dev/null || { echo 'static analysis identified user tracking library' ; false; }"
+    depends_on: "elementApp"
```
richvdh:
```suggestion
    depends_on: 
      - gplay_debug
      - gplay_release
```

richvdh (approved):

richvdh:
hrm, I see, we do not run both build steps. I have merged this, but I don't know if it will work...

richvdh:
https://buildkite.com/docs/pipelines/dependencies#order-of-operations says:

> However, if the step you're dependent on is excluded from the build due to an if condition, the dependency will be ignored and the step that depends on it will run once any other dependencies are satisfied.

so hopefully this will do the right thing

bmarty:
Thanks for the update @richvdh, I had the same thinking path as yours reading the change :)

PR #106, opened 2020-11-05
Synapse pipeline cleanup

callahad:
This reduces duplication of docker-compose files and also upgrades the docker, docker-compose, and artifacts buildkite plugins


callahad:
Would appreciate review by @matrix-org/synapse-core

anoadragon453 (commented):
Note that this will be the first time we're using `v3.7.0` of the docker-compose plugin. Also, why not update all instances of the plugin (e.g "SyTest - :python: 3.5 / :postgres: 9.6 / Workers")?

I think updating the plugin on the whole is fine. The only breaking changes in v3 of the plugin were minimal: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/releases/tag/v3.0.0

To make sure however, I've been trying to test this locally with buildkite's CLI tool and `bk local run`. Unfortunately the run requires a number of environment variables (namely `BUILDKITE_DOWNLOADS_PATH`) which are not provided by default, and attempting to provide them on the command line bears no fruit. Thus, I'd say we have a couple options:

* Merge and see what happens
* Set up a pipeline for the pipeline's repo to test PRs

I'd like to see the second one happen long-term, but am happy to try the first option for now.

callahad:
> Note that this will be the first time we're using `v3.7.0` of the docker-compose plugin. Also, why not update all instances of the plugin (e.g "SyTest - :python: 3.5 / :postgres: 9.6 / Workers")?

Those tests use the `docker` plugin, not the `docker-compose` one; figured I was just poking at Compose stuff so I limited myself to that. But it looks like the remaining plugin updates are pretty trivial / no breaking changes:

- [docker](https://github.com/buildkite-plugins/docker-buildkite-plugin/releases) 3.0.1 → 3.7.0
- [artifacts](https://github.com/buildkite-plugins/artifacts-buildkite-plugin/releases) 1.2.0 →1.3.0

I'll do that real quick.

> I'd say we have a couple options:
> 
> * Merge and see what happens
> * Set up a pipeline for the pipeline's repo to test PRs
> 
> I'd like to see the second one happen long-term, but am happy to try the first option for now.

Yeah, I'm not super happy with the CI for our CI 😉. But in this case, I'd err in the side of YOLO. Worst case, the next Synapse PR improperly fails CI and we have to roll back this commit and re-run CI. No major harm done.

callahad:
@anoadragon453 Done.

FWIW, I think we can drop the `artifacts` plugin altogether and switch to just setting `artifact_paths` [in our pipeline](https://buildkite.com/docs/pipelines/artifacts). Seems like we only need the plugin if we want one step to download artifacts created by a different job.

anoadragon453 (approved):
Yep, I'm up for YOLO as well!

> FWIW, I think we can drop the `artifacts` plugin altogether and switch to just setting `artifact_paths` [in our pipeline](https://buildkite.com/docs/pipelines/artifacts). Seems like we only need the plugin if we want one step to download artifacts created by a different job.
 
I imagine this came from cribbing off some other pipeline that needed it (matrix.org perhaps?). I would not be opposed to another PR removing it from places where it's not necessary :)

richvdh (approved):
lgtm!

PR #107, opened 2020-11-13
Synapse: Fix docker-compose variable interpolation

callahad:
...I put the vars in the wrong scope. Oops.

Signed-off-by: Dan Callahan <danc@element.io>


richvdh (approved):
lgtm

PR #108, opened 2020-11-17
Update Sytest pipeline for Dendrite

neilalexander:
This updates the Sytest pipeline so that our usual four Dendrite tests are also checked, although the result is soft-failed.

### DIFF
--- a/sytest/pipeline.yml
+++ b/sytest/pipeline.yml
@@ -185,37 +185,133 @@ steps:
         - exit_status: 2
           limit: 2
 
-  - label: "Dendrite - :go: 1.11 / :postgres: 9.6"
-    agents:
+  - label: "Dendrite - :go: 1.13 / :postgres: 9.6"
+      agents:
       queue: "medium"
-    env:
+      env:
       POSTGRES: "1"
-    command:
-      - "mkdir -p /src"
-      - "(wget -O - https://github.com/matrix-org/dendrite/archive/$BUILDKITE_BRANCH.tar.gz || wget -O - https://github.com/matrix-org/dendrite/archive/master.tar.gz) | tar -xz --strip-components=1 -C /src/"
-      - "cd /src/"
+      command:
       - "bash /bootstrap.sh dendrite"
-    plugins:
+      plugins:
       - docker#v3.0.1:
           image: "matrixdotorg/sytest-dendrite"
           propagate-environment: true
           always-pull: true
-          workdir: "/sytest"
+          workdir: "/src"
           entrypoint: '/bin/sh'
           shell: ["-x", "-c"]
           init: false
+          debug: true
           mount-buildkite-agent: false
-          mount-checkout: false
-          volumes:
-            - "./logs:/logs"
-            - ".:/sytest:ro"
+          volumes: ["./logs:/logs"]
       - artifacts#v1.2.0:
           upload: [ "logs/**/*.log", "logs/**/*.log.*", "logs/results.tap" ]
-    soft_fail:
-      - exit_status: 1
-    retry:
+      - matrix-org/annotate:
+          path: "logs/annotate.md"
+          style: "error"
+      retry:
+      soft_fail:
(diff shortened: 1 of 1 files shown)


sytest/pipeline.yml
```diff
@@ -185,37 +185,145 @@ steps:
         - exit_status: 2
           limit: 2
 
-  - label: "Dendrite - :go: 1.11 / :postgres: 9.6"
-    agents:
+  - label: "Dendrite - :go: 1.13 / :postgres: 9.6"
+      agents:
```
richvdh: this looks extremely bogus. why are you changing the indentation here?

richvdh (changes requested):

sytest/pipeline.yml
neilalexander: sigh, because VS Code is awful. That is unintentional. :/

richvdh: use a proper editor :innocent:

neilalexander: This is hopefully less bad now!

sytest/pipeline.yml
```diff
@@ -200,22 +200,130 @@ steps:
           image: "matrixdotorg/sytest-dendrite"
           propagate-environment: true
           always-pull: true
-          workdir: "/sytest"
+          workdir: "/src"
```
richvdh: I'm confused what's going on here. we run a command which attempts to construct `/src`. How does it work if we change to /src first?

sytest/pipeline.yml
```diff
@@ -200,22 +200,130 @@ steps:
...
           entrypoint: '/bin/sh'
           shell: ["-x", "-c"]
           init: false
+          debug: true
```
richvdh: is this required?

sytest/pipeline.yml
richvdh: ... and why are you making it out of sync with the synapse runs?

richvdh (changes requested):
I remain a bit confused about what your goals are here.

sytest/pipeline.yml
neilalexander: I have no idea. I just copied and pasted the pipelines from Dendrite, and `git blame` suggests Amber added it 15 months ago.

neilalexander:
I'm just trying to make it so that when we push commits to Sytest, that we also test against the four standard Dendrite test setups. I've just pretty much copied and pasted from the Dendrite `pipelines.yml` file, bar a couple of minor edits.

sytest/pipeline.yml
neilalexander: Good question, I have no idea. Also I'm not sure what you mean "out of sync"? I've just copied and pasted from the Dendrite pipelines.

neilalexander: I've put it back to `/sytest`.

sytest/pipeline.yml
richvdh: ok, it's not required. please don't cargo-cult it.

sytest/pipeline.yml
```diff
@@ -185,7 +185,7 @@ steps:
         - exit_status: 2
           limit: 2
 
-  - label: "Dendrite - :go: 1.11 / :postgres: 9.6"
+  - label: "Dendrite - :go: 1.13 / :postgres: 9.6"
```
richvdh, line 188: I'm guessing the `matrixdotorg/sytest-dendrite` image was updated a while ago and this is just updating the build name to reflect reality?

richvdh (approved):
looks sensible now.

It's worth mentioning for future reference that it's possible to use yaml entities to avoid so much duplication (see https://github.com/matrix-org/pipelines/blob/master/element-ios/pipeline.yml for example).

sytest/pipeline.yml
neilalexander: Yep indeed, we've been requiring Go 1.13 for some time now.

PR #110, opened 2020-11-24
Install the proper requirements for pysaml2 in the trial test runners.

clokep:
This installs xmlsec1 onto the Synapse trial test runners. This is [required for pysaml2](https://pysaml2.readthedocs.io/en/latest/install.html), so we need it for matrix-org/synapse#8800.

I wasn't sure if it made more sense to make a shell script in the Synapse repo and call that (like we do for old deps), but this seemed more tamper-proof.

Note that this gets installed for the old-deps build in matrix-org/synapse#8800.

I also wondered if there's a way to re-use steps or something for buildkite, but didn't see anything obvious.


richvdh:
> I also wondered if there's a way to re-use steps or something for buildkite, but didn't see anything obvious.

you can use yaml entities if you want. the element-ios pipeline has examples iirc

clokep:
> > I also wondered if there's a way to re-use steps or something for buildkite, but didn't see anything obvious.
> 
> you can use yaml entities if you want. the element-ios pipeline has examples iirc

Thanks! I made some changes based on that, not really sure how to test them though.

richvdh:
> Thanks! I made some changes based on that, not really sure how to test them though.

for future reference, there are some notes on testing pipelines at https://gitlab.matrix.org/new-vector/internal/-/wikis/BuildKite#testing-a-pipeline-locally.

richvdh (approved):
lgtm

PR #111, opened 2020-12-02
Add build and lint pipeline for complement

anoadragon453:
Adds a simple build and lint pipeline for Complement. Once this is in and works we can add Dendrite and Synapse test runs.

Note that a pipeline entry will need to be created in Buildkite as well.

@Kegsay I've added build tests for Go 1.13 and 1.15 in order to test oldest and newest supported versions of Golang. Let me know if you'd like to do something different there.

### DIFF
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -1,7 +1,7 @@
 steps:
   - command:
-      - "env"
-      - "go build -tags *"
+      - env
+      - go build -tags="*"
     label: "\U0001F528 Build / :go: 1.13"
     plugins:
       - docker#v3.7.0:
@@ -13,8 +13,8 @@ steps:
           limit: 3
 
   - command:
-      - "env"
-      - "go build -tags *"
+      - env
+      - go build -tags="*"
     label: "\U0001F528 Build / :go: 1.15"
     plugins:
       - docker#v3.7.0:
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -1,0 +1,38 @@
+steps:
+  - command:
+      - "env"
+      - "go build -tags *"
+    label: "\U0001F528 Build / :go: 1.13"
+    plugins:
+      - docker#v3.7.0:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
+    retry:
+      automatic:
+        - exit_status: 128
+          limit: 3
+
+  - command:
+      - "env"
+      - "go build -tags *"
+    label: "\U0001F528 Build / :go: 1.15"
+    plugins:
+      - docker#v3.7.0:
+          image: "golang1.15"
+          mount-buildkite-agent: false
+    retry:
+      automatic:
+        - exit_status: 128
+          limit: 3
+          
+  - command:
+      - "env"
+      # gofmt does not change its exit status depending on whether there are
+      # files in need of formatting. We have to use a slightly hacky way of
+      # checking and then printing the diff instead.
+      - "if [ -n "$(gofmt -l .)" ]; then gofmt -d && exit 1; fi"
+    label: "\U0001F9F9 Linting / :go: 1.15"
+    plugins:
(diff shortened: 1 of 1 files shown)


complement/pipeline.yml
```diff
@@ -0,0 +1,38 @@
+steps:
+  - command:
+      - "env"
```
richvdh: these are probably redundant?

richvdh (dismissed):
looks fine, but probably wants review from @Kegsay to say if these are useful build steps.

complement/pipeline.yml
anoadragon453: Probably. I cribbed it from [dendrite's pipeline](https://github.com/matrix-org/pipelines/blob/master/dendrite/pipeline.yml), and the result we get is [this](https://buildkite.com/matrix-dot-org/dendrite/builds/3908#79ed3bb4-35d5-49b4-aeeb-5af7e6e64602/48-61).

Perhaps useful to see `GOPATH` and that `GOROOT` is not set, though I don't expect any of those environment variables to ever really change.

complement/pipeline.yml
```diff
@@ -0,0 +1,38 @@
...
+      - "go build -tags *"
```
kegsay: uhh what does this do?

complement/pipeline.yml
```diff
@@ -0,0 +1,38 @@
...
+      automatic:
+        - exit_status: 128
+          limit: 3
+          
+  - command:
+      - "env"
+      # gofmt does not change its exit status depending on whether there are
+      # files in need of formatting. We have to use a slightly hacky way of
+      # checking and then printing the diff instead.
+      - "if [ -n "$(gofmt -l .)" ]; then gofmt -d && exit 1; fi"
```
kegsay: Please don't. `gofmt` shouldn't be used as part of a CI pipeline because it can absolutely change depending on the version you're running. We don't even do this on Dendrite and it hasn't bitten us.

complement/pipeline.yml
```diff
@@ -0,0 +1,38 @@
...
+    label: "\U0001F9F9 Linting / :go: 1.15"
```
kegsay: Actual linting would be welcome akin to Dendrite's.

complement/pipeline.yml
kegsay: `go build ./internal/...` would be helpful to make sure there's no weird typos in Complement such that it doesn't compile.

complement/pipeline.yml
anoadragon453: Aha, this would probably explain why it doesn't change its exit code then :)

complement/pipeline.yml
anoadragon453: Can do, that's actually even easier to crib.

complement/pipeline.yml
anoadragon453: Well, actually the `*` needs to be wrapped in quotes. But it builds all optional build tags.

Is there a reason we wouldn't want to build a specific tag in CI?

kegsay: Nope, we should build everything we can.

complement/pipeline.yml
kegsay: Indeed.

> In general, we recommend against building systems that check that source code matches the output of a specific version of gofmt. For example, a continuous integration test that fails if any code already checked into a repository is not “properly formatted” is inherently fragile and not recommended. 

https://golang.org/doc/go1.10#gofmt

anoadragon453:
Blocked on ~~https://github.com/matrix-org/complement/pull/52~~.

anoadragon453:
Added a proper linting step. Relies on ~~https://github.com/matrix-org/complement/pull/53~~.

kegsay (dismissed):

kegsay (dismissed):

richvdh (approved):

PR #114, opened 2020-12-18
Add Synapse and Dendrite runs to Complement's pipeline

anoadragon453:
Based on #111. Build steps were mostly based off of Dendrite's Complement steps: https://github.com/matrix-org/pipelines/blob/1a668cc56443ca26e31a8bb8c40130570d54f15f/dendrite/pipeline.yml#L39-L58

Adds build steps for testing Dendrite and Synapse via Complement. **Note that Synapse will fail until https://github.com/matrix-org/synapse/issues/8421 is solved (though that is quite close now).** Hopefully I can come up with a solution to that today. Review of the code should be fine, but I suggest not merging until that issue is resolved.

Another change from #111 is moving the linting step after the `- wait` so that we don't need to wait for it before running Complement tests.

### DIFF
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -60,7 +60,7 @@ steps:
            - "CI=true"
           publish: [ "8448:8448" ]
           # Complement uses Docker so pass through the docker socket. This means Complement shares
-          # the hosts Docker.
+          # the host's Docker.
           volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
     retry:
@@ -87,7 +87,7 @@ steps:
            - "CI=true"
           publish: [ "8448:8448" ]
           # Complement uses Docker so pass through the docker socket. This means Complement shares
-          # the hosts Docker.
+          # the host's Docker.
           volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
     retry:
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -69,7 +69,7 @@ steps:
 
   - command:
       # Build the Dendrite for Complement docker image
-      - "cd /src && docker build -t complement-dendrite -f dockerfiles/Dendrite.Dockerfile ."
+      - "docker build -t complement-dendrite -f dockerfiles/Dendrite.Dockerfile ."
       # Run the tests!
       - "COMPLEMENT_BASE_IMAGE=complement-dendrite go test -v -tags "msc2836" ./tests"
     label: "\U0001F9EA Complement / Dendrite Monolith / :go: 1.15"
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -49,10 +49,7 @@ steps:
       - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags "synapse_blacklist" ./tests
     label: "\U0001F9EA Complement / Synapse Monolith / :go: 1.15"
     agents:
-      # As we provide access to the host docker socket below, use an ephemeral agent which
-      # will be destroyed after this job is complete. This will prevent any potentially
-      # malicious code running in this build agent from tampering with any further jobs.
-      queue: "medium-ephemeral"
+      queue: "medium"
     plugins:
       - docker#v3.7.0:
           # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
@@ -79,10 +76,7 @@ steps:
       - COMPLEMENT_BASE_IMAGE=complement-dendrite go test -v -tags "msc2836" ./tests
     label: "\U0001F9EA Complement / Dendrite Monolith / :go: 1.15"
     agents:
-      # As we provide access to the host docker socket below, use an ephemeral agent which
-      # will be destroyed after this job is complete. This will prevent any potentially
-      # malicious code running in this build agent from tampering with any further jobs.
-      queue: "medium-ephemeral"
+      queue: "medium"
     plugins:
       - docker#v3.7.0:
           # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -42,13 +42,8 @@ steps:
           limit: 3
 
   - command:
-      # Download the latest checkout of Synapse develop
-      - "wget -N https://github.com/matrix-org/synapse/archive/develop.tar.gz -O synapse.tar.gz && tar -xzf synapse.tar.gz && cd synapse"
-      # Build the Synapse production docker image
-      - docker build -t matrixdotorg/synapse -f docker/Dockerfile .
-      # Now hop over to complement's checkout and build the Complement-specific Synapse docker image,
-      # which inherits from the production Synapse Dockerfile
-      - "cd /src && docker build -t complement-synapse -f dockerfiles/Synapse.Dockerfile ."
+      # Build the Synapse for Complement docker image, which inherits from the latest Synapse release image on Dockerhub
+      - "docker build -t complement-synapse -f dockerfiles/Synapse.Dockerfile ."
       # Run the tests!
       - "COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags "synapse_blacklist,msc2403" ./tests"
     label: "\U0001F9EA Complement / Synapse Monolith / :go: 1.15"
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -49,7 +49,10 @@ steps:
       - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags "synapse_blacklist" ./tests
     label: "\U0001F9EA Complement / Synapse Monolith / :go: 1.15"
     agents:
-      queue: "medium"
+      # As we provide access to the host docker socket below, use an ephemeral agent which
+      # will be destroyed after this job is complete. This will prevent any potentially
+      # malicious code running in this build agent from tampering with any further jobs.
+      queue: "medium-ephemeral"
     plugins:
       - docker#v3.7.0:
           # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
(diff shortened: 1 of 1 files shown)


complement/pipeline.yml
```diff
@@ -37,3 +40,60 @@ steps:
       automatic:
         - exit_status: 128
           limit: 3
+
+  - command:
+      # Download the latest checkout of Synapse develop
+      - "wget -N https://github.com/matrix-org/synapse/archive/develop.tar.gz -O synapse.tar.gz && tar -xzf synapse.tar.gz && cd synapse"
```
kegsay: Please use a stable release e.g `master`. We aren't interested in bleeding edge Synapse stuff here, we care about Complement instead.

complement/pipeline.yml
```diff
@@ -37,3 +40,60 @@ steps:
...
+      # Build the Synapse production docker image
+      - docker build -t matrixdotorg/synapse -f docker/Dockerfile .
```
kegsay: We really have to do this and can't use the docker hub version?

complement/pipeline.yml
```diff
@@ -37,3 +40,60 @@ steps:
...
+      queue: "medium"
+    plugins:
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          # Complement needs to know if it is running under CI
+          environment:
+           - "CI=true"
+          publish: [ "8448:8448" ]
```
kegsay, line 93: Have you tried running this pipeline locally? I doubt this will work because synapse/dendrite will be fighting for the same ports.

kegsay (changes requested):

complement/pipeline.yml
anoadragon453: Good point. We'll be testing `develop` in Synapse's pipeline anyways.

Heads up that building Complement's [Dendrite.Dockerfile](https://github.com/matrix-org/complement/blob/master/dockerfiles/Dendrite.Dockerfile) will build Dendrite master however. We may want to change that to download the latest release tag instead?

complement/pipeline.yml
anoadragon453: Now that we're using the latest Synapse release we can totally use dockerhub!

complement/pipeline.yml
anoadragon453: I did not, and thank you for the suggestion, I discovered a few bugs along the way! I also wrote some documentation on running pipelines locally: https://github.com/matrix-org/pipelines/pull/116 If you often do so yourself, can you take a look and see if what I was doing is overkill?

Otherwise I didn't experience any issues with port allocation during my test runs. Both Synapse and Dendrite are created in separate build steps, so they shouldn't interfere with one another?

anoadragon453:
This pipeline passes when run locally.

complement/pipeline.yml
```diff
@@ -37,3 +40,57 @@ steps:
...
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the hosts Docker.
```
richvdh:
```suggestion
          # the host's Docker.
```

complement/pipeline.yml
```diff
@@ -37,3 +40,57 @@ steps:
...
```
richvdh:
```suggestion
          # the host's Docker.
```

complement/pipeline.yml
```diff
@@ -37,3 +40,57 @@ steps:
...
+      - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags "synapse_blacklist" ./tests
+    label: "\U0001F9EA Complement / Synapse Monolith / :go: 1.15"
+    agents:
+      queue: "medium"
+    plugins:
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          # Complement needs to know if it is running under CI
```
richvdh, line 63: I can't help but feel this is a terrible smell. The whole point of CI is to test that the thing works correctly, so why do we have different code paths under CI?

complement/pipeline.yml
```diff
@@ -37,3 +40,57 @@ steps:
...
+          environment:
+           - "CI=true"
+          publish: [ "8448:8448" ]
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the hosts Docker.
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
```
richvdh: this will give access to run arbitrary executables on the build server, so will mean that we can't enable CI for third-party PRs. possibly we could allow it if we ran the build on one of the burn-after-build agents, but we should probably discuss that in CI internal first.

complement/pipeline.yml
richvdh: it'll presumably fail if you try to run both buildsteps on the same machine, but none of our build agents are set up to do that, so you can probably get away with it.

richvdh (commented):
looks mostly fine from the buildkite POV, but would appreciate kegsay's review of the complement bits.

complement/pipeline.yml
anoadragon453: I ran the entire pipeline solely on my machine though.

richvdh: sorry, what I mean is if you run both build steps *concurrently* on one machine, either by configuring the buildkite agent process to run multiple build steps at once, or by running multiple buildkite agent processes. 

Presumably you did neither of the above, but what I'm saying is, that's fine. None of the buildkite agents we use in our "production" buildkite do so either.

complement/pipeline.yml
anoadragon453: I agree that this could be used to diverge CI test runs from what's intended during development if this environment variable is abused. And we'll need to ensure this doesn't happen.

However at the moment the purpose of this variable is very simple. Some changes are needed if Complement itself is running in a docker container (and given access to the host's docker) versus running on the host directly. See where it is used:

Talking to the right docker runtime:

https://github.com/matrix-org/complement/blob/ee3715d9a2788263ac00990c260b2cf8a4f168ed/internal/docker/builder.go#L51-L56

Placement of CA files:

https://github.com/matrix-org/complement/blob/ee3715d9a2788263ac00990c260b2cf8a4f168ed/internal/docker/builder.go#L343-L408

https://github.com/matrix-org/complement/blob/ee3715d9a2788263ac00990c260b2cf8a4f168ed/internal/federation/server.go#L228-L245

This could possibly be implemented in a different manner than an environment variable, but I think the need for split code paths would remain.

complement/pipeline.yml
anoadragon453: We had a quick chat internally and have opted for using burn-after-build agents, which needed only a quick change made in cf896c8fd797130d60733754821790b90d0ab585.

complement/pipeline.yml
anoadragon453: Presumably this concern may be further eliminated by now using burn-after-build agents.

complement/pipeline.yml
richvdh: thanks, that's helpful.

Looking at this harder, I think the problem is really just that it's a crappy name for the env var. It's not "are we running under CI", but rather "is complement running under docker?" Presumably that's something you could detect (cf https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker), but just renaming `CI` to `COMPLEMENT_IN_DOCKER` or something would be a good start.

To be clear: I'm not insisting you change it - I'm just saying it might be a thing to improve in the future.

complement/pipeline.yml
```diff
@@ -37,3 +40,63 @@ steps:
...
+      # https://github.com/matrix-org/complement/blob/master/dockerfiles/Synapse.Dockerfile.
+      - docker build -t complement-synapse -f dockerfiles/Synapse.Dockerfile dockerfiles/
+      # Run the tests!
+      - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags "synapse_blacklist" ./tests
+    label: "\U0001F9EA Complement / Synapse Monolith / :go: 1.15"
+    agents:
+      # As we provide access to the host docker socket below, use an ephemeral agent which
+      # will be destroyed after this job is complete. This will prevent any potentially
+      # malicious code running in this build agent from tampering with any further jobs.
+      queue: "medium-ephemeral"
```
richvdh: shower thought on this: even the `ephemeral` build agents still have the buildkite agent token, which gives you the ability to run arbitrary jobs on *other* agents, so this isn't going to cut it. I think we'll have to lock the build down to trusted branches, which means you may as well go back to `medium`, since they are faster.

sorry for the runaround.

richvdh (dismissed):

complement/pipeline.yml
kegsay: Hmm, probably yeah.

kegsay (dismissed):

complement/pipeline.yml
anoadragon453: Good points. I've created an issue for this in Complement's repo: https://github.com/matrix-org/complement/issues/60

Suggest we do so after merging this.

complement/pipeline.yml
anoadragon453: Ah, darn OK. Have removed them then. Be sure to turn off Contributor builds as well.

richvdh: we should probably stick a warning in at the top of the file to remind ourselves not to enable contrib PRs, and why:

```
# WARNING: do not enable builds of third-party PRs on this pipeline. The pipeline is configured to
# give access to the Docker socket to the Complement test binaries, so a malicious fork of
# Complement could start a Docker container which extracts the Buildkite agent token and hence
# compromises this and other Buildkite agents.
```

kegsay: Looking at https://buildkite.com/docs/pipelines/environment-variables#environment-variable-precedence can't we just clobber the agent token in this pipeline by setting some bogus value?

richvdh: for this pipeline, sure, but we still have the problem that the token exists somewhere on the parent host, and we are giving unrestricted access to the parent host via the docker socket.

anoadragon453: Have added the warning.

richvdh (approved):
ok cool, I'm going to merge this with contrib PRs disabled. If we want to split it into two pipelines in future, that can be a separate PR.

PR #116, opened 2020-12-29
Add instructions on how to run pipelines locally

anoadragon453:
I wanted to test #114 locally, and found out a working method to do so. I'd be a little surprised if this were the most direct method however, as it requires retrieving tokens from buildkite and creating an organisation just to run some tests locally.

I found that attempting to run `bk local run /path/to/pipeline.yml` would fail unless `buildkite-agent` was running though, and `buildkite-agent` refused to start up without a valid agent token, so ¯\\\_(ツ)_/¯.

I'm hoping that this won't mean a bunch of random organisations get littered over someone's dashboard somewhere though, or whether we should set up a single, testing organisation for this purpose instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Buildkite Pipelines for Matrix.org Projects.
 See the intructions at
 https://gitlab.matrix.org/new-vector/internal/-/wikis/BuildKite#testing-a-pipeline-locally.
 
-This is an internal page. If you are not allowed to view this page, know that
+The above is an internal page. If you are not allowed to view this page, know that
 the instructions effectively boil down to:
 
 1. Install [buildkite-agent](https://github.com/buildkite/agent) and [buildkite-cli](https://github.com/buildkite/cli).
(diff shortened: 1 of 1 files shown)


richvdh (changes requested):
It would probably be better to link to https://gitlab.matrix.org/new-vector/internal/-/wikis/BuildKite#testing-a-pipeline-locally rather than having a separate set of instructions.

`bk local run` should start a local buildkite agent with a temporary agent token. Is `buildkite-agent` on your path?

anoadragon453:
> `bk local run` should start a local buildkite agent with a temporary agent token. Is `buildkite-agent` on your path?

Indeed, and I believe I saw somewhere that `bk run` was an alias to `bk local run`, but I might've missed something. I'll take another look and revise the process soon.

Another reason is that one only gets an organisation for free for ~2 weeks, so it's not the best solution :yum:

richvdh:
I'm going to close this for now pending updates.

anoadragon453:
@richvdh Thanks. I used the instructions on the wiki to great success. I'm not sure why a similar set of commands failed so spectacularly the first time round.

I've updated this PR to link to the instructions, with a little summary below.

richvdh (approved):
lgtm. I wonder if we should just link in the other direction, to save duplication, but :man_shrugging:

PR #117, opened 2020-12-31
Use blacklist and remove soft-fail status for Synapse's Complement run

anoadragon453:
Once https://github.com/matrix-org/synapse/issues/8421 is sorted (one outstanding PR) Synapse should now pass all generally applicable tests in Complement. At this point, we'll want to no longer soft-fail Complement CI runs so that we'll be alerted if we break anything.

This PR also sets Synapse's Complement blacklist, which was introduced in https://github.com/matrix-org/complement/pull/55.

Mergable once https://github.com/matrix-org/pipelines/pull/114 is in and Complement's CI shows Synapse as green.

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -577,9 +577,6 @@ steps:
     label: "\U0001F9EA Complement"
     agents:
       queue: "medium"
-    # Continue even on failure for now
-    soft_fail:
-      - exit_status: 1
     plugins:
       - docker#v3.7.0:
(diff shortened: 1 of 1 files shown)


anoadragon453:
> Mergable once #114 is in and Complement's CI shows Synapse as green.

[This is now the case!](https://buildkite.com/matrix-dot-org/complement/builds/23)

richvdh (approved):
:tada:

PR #118, opened 2021-01-06
Synapse: add a build step to run with no optional deps

richvdh:
requires https://github.com/matrix-org/synapse/pull/9030

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -10,6 +10,13 @@ x-yaml-aliases:
       apt-get update && apt-get install -y xmlsec1
       python -m pip install tox
 
+  retry: &retry_setup
+    automatic:
+      - exit_status: -1
+        limit: 2
+      - exit_status: 2
+        limit: 2    
+    
 env:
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -10,6 +10,13 @@ x-yaml-aliases:
       apt-get update && apt-get install -y xmlsec1
       python -m pip install tox
 
+  retry: &retry_setup
+    automatic:
+      - exit_status: -1
+        limit: 2
+      - exit_status: 2
+        limit: 2    
+
```
richvdh, line 19: thought I might as well factor this out rather than add yet another copy.

synapse/pipeline.yml
```diff
@@ -186,13 +178,24 @@ steps:
...
 #          parallel: "true"
-    retry:
-      automatic:
-        - exit_status: -1
-          limit: 2
-        - exit_status: 2
-          limit: 2
+    retry: *retry_setup
 
+  - label: ":python: 3.8 / noextras / SQLite"
```
richvdh, line 183: why py 3.8? no idea. it was chosen pretty arbitrarily.

erikjohnston (approved):

PR #119, opened 2021-01-07
Update Synapse pipelines for Python 3.9

### LINKED ISSUE #104: Add Python 3.9 to Synapse tests
Synapse supports Python 3.9.

We should probably have some tests to make sure it stays that way.

Right now our [pipeline](https://github.com/matrix-org/pipelines/blob/master/synapse/pipeline.yml) yields a test matrix like this:

__Twisted Trial:__

|               | 3.5 | 3.6 | 3.7 | 3.8 | 3.9 |
| ------------- | :-: | :-: | :-: | :-: | :-: |
| Sqlite        | ✅¹ |  ✅  |  ✅  |     |
| Postgres 9.5  |  ✅  |     |     |     |
| Postgres 9.6  |     |     |     |     |
| Postgres 10 |     |     |     |     |
| Postgres 11 |     |     |  ✅  |     |
| Postgres 12 |     |     |     |  ✅  |
| Postgres 13 |     |     |     |     |

¹: Also includes an additional `olddeps` variant

__Sytest:__

|               | 3.5  | 3.6 | 3.7 | 3.8 | 3.9 |
| ------------- | :--: | :-: | :-: | :-: | :-: |
| Sqlite        |  💧  |     |     |     |
| Postgres 9.5  |      |     |     |     |
| Postgres 9.6  | 💧💦 |     |     |     |
| Postgres 10 |     |     |     |     |
| Postgres 11 |      |     | 💦²  |     |
| Postgres 12 |      |     |     | 💧  |
| Postgres 13 |     |     |     |     |

💧: Monolith
💦: Multiple Workers
²: Also includes an additional `with redis` variant

__Synapse Port DB:__

|               | 3.5 | 3.6 | 3.7 | 3.8 | 3.9 |
| ------------- | :-: | :-: | :-: | :-: | :-: |
| Sqlite        |     |     |     |     |
| Postgres 9.5  |  ✅  |     |     |     |
| Postgres 9.6  |     |     |     |     |
| Postgres 10 |     |     |     |     |
| Postgres 11 |     |     |  ✅  |     |
| Postgres 12 |     |     |     |     |
| Postgres 13 |     |     |     |     |

This raises all kinds of questions (Do we really need all of those Postgres versions? And why _those_ specific ones, when upstream is up to 13.0?)... but more immediately, where should Python 3.9 fit in here?

Seems like there are holes for Twisted Trial runs with:

- Py 3.8 + Sqlite
- Py 3.9 + Sqlite
- Py 3.9 + Postgres ?.?

But what about Sytest?

callahad:
Fixes #104

Reviewable commit-by-commit.

This goes a bit further than a straight minimal update to Python 3.9; changes could be broken out separately if desired.

Specifically, I try to articulate *what* we're hoping to test in each section, then add / remove / re-order tests to align with that description.

I also re-named several tests to hopefully avoid truncating test names on the Buildkite dashboard, which I've found can obscure information, especially around the sytest and port_db suites.

If all goes well, we should have something like this:

__Before:__

Buildkite status

__After:__

Buildkite status

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -294,9 +294,9 @@ steps:
   # We mostly care about testing each topology.
   # To vary the Python and Postgres versions, we use Docker images which are based
   # on an assortment of Debian releases.
-  #   - "py35" is Debian Stretch (Python 3.5, Postgres 9.6)
-  #   - "py37" is Debian Buster (Python 3.7, Postgres 11)
-  #   - "py38" is Debian Bullseye (Python 3.9, Postgres 13)
+  #   - "py35" is Debian  9 Stretch (Python 3.5, Postgres 9.6)
+  #   - "py37" is Debian 10 Buster (Python 3.7, Postgres 11)
+  #   - "py38" is Debian 11 Bullseye (Python 3.9, Postgres 13)
   #
   # Our intent is to test:
   #   - Monolith:
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -335,7 +335,7 @@ steps:
           style: "error"
     retry: *retry_setup
 
-  - label: "SyTest Monolith :postgres: :debian 9:"
+  - label: "SyTest Monolith :postgres: :debian: 9"
     agents:
       queue: "medium"
     env:
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -259,39 +259,16 @@ steps:
           upload: [ "_trial_temp/*/*.log" ]
     retry: *retry_setup
 
-  - label: ":python: 3.7 / :postgres: 11"
+  - label: ":python: 3.9 / :postgres: 13"
     agents:
       queue: "medium"
     env:
       TRIAL_FLAGS: "-j 8"
-      PYTHON_VERSION: "3.7"
-      POSTGRES_VERSION: "11"
+      PYTHON_VERSION: "3.9"
+      POSTGRES_VERSION: "13"
     command:
       - *trial_setup
-      - "python -m tox -e py37-postgres,combine"
-    plugins:
-      - matrix-org/download#v1.1.0:
-          urls:
-            - https://raw.githubusercontent.com/matrix-org/pipelines/master/synapse/docker-compose.yaml
-            - https://raw.githubusercontent.com/matrix-org/pipelines/master/synapse/docker-compose-env
-      - docker-compose#v3.7.0:
-          run: testenv
-          config:
-            - /tmp/download-${BUILDKITE_BUILD_ID}/docker-compose.yaml
-      - artifacts#v1.3.0:
-          upload: [ "_trial_temp/*/*.log" ]
-    retry: *retry_setup
-
-  - label: ":python: 3.8 / :postgres: 12"
-    agents:
-      queue: "medium"
-    env:
-      TRIAL_FLAGS: "-j 8"
-      PYTHON_VERSION: "3.8"
-      POSTGRES_VERSION: "12"
-    command:
-      - *trial_setup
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -299,11 +333,9 @@ steps:
       - matrix-org/annotate:
           path: "logs/annotate.md"
           style: "error"
-#      - matrix-org/coveralls#v1.0:
-#          parallel: "true"
     retry: *retry_setup
 
-  - label: "SyTest - :python: 3.5 / :postgres: 9.6 / Monolith"
+  - label: "SyTest Monolith :postgres: :debian 9:"
```
```suggestion
  - label: "SyTest Monolith :postgres: :debian: 9"
```

synapse/pipeline.yml
```diff
@@ -267,17 +280,38 @@ steps:
...
+  #   1. Topology (Monolith, Workers, Workers w/ Redis)
+  #   2. Database (SQLite or PostgreSQL)
+  #   3. Python Version
+  #
+  # We mostly care about testing each topology.
+  # To vary the Python and Postgres versions, we use Docker images which are based
+  # on an assortment of Debian releases.
+  #   - "py35" is Debian Stretch (Python 3.5, Postgres 9.6)
+  #   - "py37" is Debian Buster (Python 3.7, Postgres 11)
+  #   - "py38" is Debian Bullseye (Python 3.9, Postgres 13)
```
richvdh: probably worth noting the debian release numbers here, since we refer to them as 8/9/10 below.

synapse/pipeline.yml
```diff
@@ -267,17 +280,38 @@ steps:
...
-#          parallel: "true"
     retry: *retry_setup
 
   ################################################################################
   #
   # Sytest
   #
+  # Our tests have three dimensions:
+  #   1. Topology (Monolith, Workers, Workers w/ Redis)
+  #   2. Database (SQLite or PostgreSQL)
```
richvdh: also, we can run against *multiple* databases (`MULTI_POSTGRES`=1).

richvdh (commented):
seems generally sensible to me. a couple of thoughts.

synapse/pipeline.yml
callahad: Ah, good point. Our existing tests cover:

- Single Postgres (Monolith, Workers w/ Redis)
- Multi Postgres (Workers)

We do not cover:

- Single Postgres (Workers)
- Multi Postgres (Monolith, Workers w/ Redis)

I'll update the comment accordingly, but since this is the status quo I'll defer changing anything for a separate discussion and PR.

callahad: Marking as resolved

synapse/pipeline.yml
callahad: Done in ed3a4ad48c26dc8ac965f2026653958eea87a750

callahad:
I think this is good to go (note I've not actually _tested_ this on Buildkite, since that's non-trivial, but there shouldn't be any risky changes.)

synapse/pipeline.yml
```diff
@@ -288,24 +288,27 @@ steps:
   #
   # Our tests have three dimensions:
   #   1. Topology (Monolith, Workers, Workers w/ Redis)
-  #   2. Database (SQLite or PostgreSQL)
+  #   2. Database (SQLite, PostgreSQL)
   #   3. Python Version
   #
+  # Tests can run against either a single or multiple PostgreSQL databases.
```
richvdh, line 294: technically Synapse can run against a sqlite and a postgres, though doing that seems slightly insane and the sytest setup scripts have no support for such a thing.

richvdh (approved):

PR #121, opened 2021-01-14
Ensure Hebrew translation will be copied to the correct folder

bmarty:
Same pb than for Indonesian translation, but for Hebrew

### DIFF
--- a/element-android/pipeline.yml
+++ b/element-android/pipeline.yml
@@ -123,17 +123,19 @@ steps:
           image: "openjdk"
           mount-buildkite-agent: false
 
-  # Check that indonesians files are identical.
-  # Due to Android issue, the resource folder must be values-in/, and Weblate export data into values-id/.
-  # If this step fails, it means that Weblate has updated the file in value-id/ so to fix it, copy the file to values-in/
-  - label: "Indonesian"
+  # Check that some language files are identical.
+  # Due to Android issue, some of the resource folders, used at runtime, do not have the same name than those used by Weblate export
+  # - Indonesian:
+  # must be values-in/, and Weblate export data into values-id/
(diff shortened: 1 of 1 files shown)


element-android/pipeline.yml
```diff
@@ -130,6 +130,12 @@ steps:
     command:
       - "diff ./vector/src/main/res/values-id/strings.xml ./vector/src/main/res/values-in/strings.xml"
 
+  # Same thing for Hebrew
+  # If this step fails, it means that Weblate has updated the file in value-he/ so to fix it, copy the file to values-iw/
+  - label: "Hebrew"
+    command:
```
richvdh: it might be better to merge these steps together, so that we don't need to wait for two build agents to start and check out the code?

```suggestion
```

(you'll need to update the `label` as well)

richvdh (commented):
Some other thoughts:

Could the underlying issue be solved with a symlink between the two files?

Otherwise it might be better to create a script in the repo that we can run from the pipeline, so it is easier to add new languages?

bmarty:
> Otherwise it might be better to create a script in the repo that we can run from the pipeline, so it is easier to add new languages?

I've though about that, but I think it will become a security issue, no? That's why pipeline are in another repository.

richvdh:
> I've though about that, but I think it will become a security issue, no? That's why pipeline are in another repository.

we can run the script under docker, same as the other build steps.

element-android/pipeline.yml
bmarty: Ok, I will merge those 2 steps.
The (small)  problem is that if the first diff fails, we will not be able to see if the second one will fail also, because it will not be executed.

richvdh (approved):
lgtm

PR #122, opened 2021-01-21
Synapse: Add a step for running Complement tests in worker mode

anoadragon453:
This PR extends the Complement section of Synapse's pipeline to run Complement tests under worker mode. This PR is ready for review, but should not be merged until the relevant [Synapse](https://github.com/matrix-org/synapse/pull/9162) and [Complement](https://github.com/matrix-org/complement/pull/62) PRs are.

## How does this work?

We start with the official Synapse docker image. This creates a python environment with all of Synapse's dependencies installed. Then we make use of the new `Dockerfile-workers` to install supervisord, redis and nginx. Redis is used for inter-process communication, nginx is used for routing network requests to each worker and supervisord will start and maintain each process.

Finally, we use the `SynapseWorkers` dockerfile in Complement to install postgresql into the container, set up caddy server to easily take advantage of Complement's CA, inject some Synapse options to disable rate-limiting, enable registration etc. and then finally run a script which generates config files supervisord, nginx and Synapse depending on which workers you want to run.

### DIFF
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -27,6 +27,9 @@ x-yaml-aliases:
         volumes:
           - "/var/run/docker.sock:/var/run/docker.sock"
 
+    # The common command for running Complement tests against Synapse
+    - &complement_run_tests_command "go test -v -tags synapse_blacklist,msc2946,msc3083 ./tests"
+
   retry: &retry_setup
     automatic:
       - exit_status: -1
@@ -590,10 +593,12 @@ steps:
       # Build a second docker image on top of the above image. This one sets up Synapse with a generated config file,
       # signing and SSL keys so Synapse can run and federate
       - "docker build -t complement-synapse -f complement-master/dockerfiles/Synapse.Dockerfile complement-master/dockerfiles"
+      - "export COMPLEMENT_BASE_IMAGE=complement-synapse"
 
       # Finally, compile and run the tests.
       - "cd complement-master"
-      - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v -tags synapse_blacklist,msc2946,msc3083 ./tests"
+      - *complement_run_tests_command
+
     label: "\U0001F9EA Complement | Synapse Monolith"
     agents:
       queue: "medium"
@@ -619,11 +624,12 @@ steps:
       - "tar -xzf master.tar.gz"
       # Build an image for running the above worker setup in Complement. This involves
       # disabling rate-limiting, using Complement's CA etc.
-      - "docker build -t complement-synapse -f complement-master/dockerfiles/SynapseWorkers.Dockerfile complement-master/dockerfiles"
+      - "docker build -t complement-synapse-workers -f complement-master/dockerfiles/SynapseWorkers.Dockerfile complement-master/dockerfiles"
+      - "export COMPLEMENT_BASE_IMAGE=complement-synapse-workers"
 
       # Finally, compile and run the tests.
       - "cd complement-master"
-      - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v -tags synapse_blacklist,msc2946,msc3083 ./tests"
+      - *complement_run_tests_command
     label: "\U0001F9EA Complement | Synapse Workers"
     agents:
       # Running every worker takes a beefy system
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -21,8 +21,6 @@ x-yaml-aliases:
           - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest"
           # Complement needs to know if it is running under CI
           - "CI=true"
-          # Enable debug logging
-          - "COMPLEMENT_DEBUG=1"
         publish: [ "8448:8448" ]
         # Complement uses Docker so pass through the docker socket. This means Complement shares
         # the hosts Docker.
--- a/synapse/pipeline.yml
+++ b/synapse/pipeline.yml
@@ -545,7 +545,7 @@ steps:
       # Build a docker image from the checked out Synapse source
       - "docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile ."
       # We use the complement:latest image to provide Complement's dependencies, but want
-      # to actually run against the latest version of Complement, so download it here.
+      # to actually run against the master branch of Complement, so download it here.
       - "wget https://github.com/matrix-org/complement/archive/master.tar.gz"
       - "tar -xzf master.tar.gz"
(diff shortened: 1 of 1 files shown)


synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+      # We use the complement:latest image to provide Complement's dependencies, but want
+      # to actually run against the master branch of Complement, so download it here.
+      - "wget https://github.com/matrix-org/complement/archive/master.tar.gz"
+      - "tar -xzf master.tar.gz"
+      # Build an image for running the above worker setup in Complement. This involves
+      # disabling rate-limiting, using Complement's CA etc.
+      - "docker build -t complement-synapse -f complement-master/dockerfiles/SynapseWorkers.Dockerfile complement-master/dockerfiles"
+      # Finally, compile and run the tests.
+      - "cd complement-master"
+      - "COMPLEMENT_BASE_IMAGE=complement-synapse:latest go test -v -tags synapse_blacklist ./tests"
```
richvdh: my first reaction to this was "why can't we stick this stuff in a script in the repo" - to which the answer is "this stuff runs in a privileged docker container, so we can't open it up to community PRs". It would be worth sticking a comment in to that effect.

synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          environment:
+            # Complement needs to know if it is running under CI
+            - "CI=true"
+            # Enable debug logging
+            - "COMPLEMENT_DEBUG=1"
+            # Enabled Complement's certificate authority for authentication of
```
```suggestion
            # Enable Complement's certificate authority for authentication of
```

synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+      # Running every worker takes a beefy system
+      queue: "xlarge"
+    plugins:
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          environment:
+            # Complement needs to know if it is running under CI
+            - "CI=true"
```
richvdh: :roll_eyes:

synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+            # Enable debug logging
+            - "COMPLEMENT_DEBUG=1"
```
richvdh: is there a plan to turn this off again, or?

synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+            # Enabled Complement's certificate authority for authentication of
+            # federation requests
+            - "COMPLEMENT_CA=true"
```
richvdh: does this *need* to be optional?

synapse/pipeline.yml
```diff
@@ -570,3 +570,46 @@ steps:
...
+            # Starting up all the processes in this container can take a little while.
+            # The default is 200. Let's wait a bit longer to decrease the chance of a
+            # startup timeout
+            - "COMPLEMENT_VERSION_CHECK_ITERATIONS=500"
+          publish: [ "8448:8448" ]
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the hosts Docker.
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
```
richvdh: It would be nice to make use of yaml entities so that the bits that are shared between the complement runs (the docker-plugin settings, and some of the `commands`) don't have to be C&Ped.

richvdh (changes requested):

synapse/pipeline.yml
anoadragon453: https://github.com/matrix-org/complement/issues/60

I'm planning to change that myself once these PRs are in :)

synapse/pipeline.yml
anoadragon453: Hm, well this parameter will log every request sent to the homeserver with its contents, which I'd find is useful for debugging. It does also print debug logs from Complement, which at the moment are not very much, but may become noisy in the future.

Do we want all client requests to the homeserver logged in CI? Or is that too noisy, and people can re-run the tests locally/check the test code if they'd like to see the exact requests being made?

There's also a separate discussion about potentially splitting this into two environment variables. One for printing client requests, and the other for Complement debug logs.

synapse/pipeline.yml
anoadragon453: Possibly not, though existing images expecting to fill the `/ca` directory with their own files (for example, [Synapse Monolith's](https://github.com/matrix-org/complement/blob/df150033eea5815c41111f700c5c00b9eb624f25/dockerfiles/Synapse.Dockerfile#L19)) will need to be updated if this option is on by default.

I've created an issue for discussion on the Complement repo here: https://github.com/matrix-org/complement/issues/65

synapse/pipeline.yml
richvdh: it sounds rather noisy to me. We don't do that for sytest.

richvdh: (alternatively, if it's useful in CI, why wouldn't it be useful always?)

synapse/pipeline.yml
anoadragon453: I was able to reduce the `plugins` section quite significantly, but didn't see as clear a path with `commands:` as there's some differences and commonalities interspersed with one another.

richvdh: looks the same apart from the name of the dockerfile? so have two yaml entities, one which does the bits before that, and one which does the bits after?

or even: stick the name of a dockerfile in an env var and set it somewhere else in the step config?

synapse/pipeline.yml
anoadragon453: Yeah, I think the intention is that it's useful for Complement's CI, but not Synapse's. If we need request logs on failing tests, then we can advocate for that to work outside of the `COMPLEMENT_DEBUG` env var.

I'll remove it.

synapse/pipeline.yml
anoadragon453: It seems as though environment variables specified in `environment` aren't used when launching commands specified in `commands`. The environment variables that seem to work there are those found in [docker-compose-env](https://github.com/matrix-org/pipelines/blob/master/synapse/docker-compose-env). But that doesn't really help us much.

richvdh:
closing this, because synapse has moved its CI to GHA.

PR #123, opened 2021-01-21
Complement: Add Synapse Workers build step

anoadragon453:
This PR extends the Complement pipeline to run tests on Synapse in worker mode. This PR is ready for review, but should not be merged until the relevant [Synapse](https://github.com/matrix-org/synapse/pull/9162) and [Complement](https://github.com/matrix-org/complement/pull/62) PRs are.

I have confirmed that this pipeline succeeds locally.

### DIFF
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -74,6 +74,53 @@ steps:
           limit: 3
 
   - command:
+      # Build the Synapse Workers docker image, which is located at:
+      # https://github.com/matrix-org/synapse/blob/master/docker/Docker-workers
+      # Note that this image is based on matrixdotorg/synapse:latest, which we're just
+      # pulling from docker hub
+      - wget https://github.com/matrix-org/synapse/archive/master.tar.gz
+      - tar -xzf master.tar.gz
+      - docker build -t matrixdotorg/synapse:workers -f synapse-master/docker/Dockerfile-workers synapse-master/
+      # Build the accompanying Synapse workers Complement docker container. This specifically
+      # sets up a Synapse worker setup for Complement, and is located at:
+      # https://github.com/matrix-org/complement/blob/master/dockerfiles/SynapseWorkers.Dockerfile.
+      - docker build -t complement-synapse -f dockerfiles/SynapseWorkers.Dockerfile dockerfiles/
+      # Run the tests!
+      - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags synapse_blacklist ./tests
+    label: "\U0001F9EA Complement / Synapse Workers / :go: 1.15"
+    agents:
+      # Running every worker takes a beefy system
+      queue: "xlarge"
+    plugins:
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          # Complement needs to know if it is running under CI
+          environment:
+            # Complement needs to know if it is running under CI
+            - "CI=true"
+            # Enable debug logging
+            - "COMPLEMENT_DEBUG=1"
+            # Enabled Complement's certificate authority for authentication of
+            # federation requests
+            - "COMPLEMENT_CA=true"
+            # Starting up all the processes in this container can take a little while.
+            # The default is 200. Let's wait a bit longer to decrease the chance of a
+            # startup timeout
+            - "COMPLEMENT_VERSION_CHECK_ITERATIONS=500"
+          publish: [ "8448:8448" ]
+          # Complement uses Docker so pass through the docker socket. This means Complement shares
+          # the host's Docker.
+          volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
+    retry:
+      automatic:
+        - exit_status: 128
+          limit: 3
+
+
+  - command:
       # Build the Dendrite for Complement docker image, which is located at:
       # https://github.com/matrix-org/complement/blob/master/dockerfiles/Dendrite.Dockerfile.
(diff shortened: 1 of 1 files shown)


kegsay (dismissed):

complement/pipeline.yml
```diff
@@ -73,6 +73,53 @@ steps:
         - exit_status: 128
           limit: 3
 
+  - command:
+      # Build the Synapse Workers docker image, which is located at:
+      # https://github.com/matrix-org/synapse/blob/master/docker/Docker-workers
+      # Note that this image is based on matrixdotorg/synapse:latest, which we're just
+      # pulling from docker hub
+      - wget https://github.com/matrix-org/synapse/archive/master.tar.gz
+      - tar -xzf master.tar.gz
```
richvdh, line 84: do we really need the whole of the synapse repo? If it's just the dockerfile you're after, just curl that?

complement/pipeline.yml
```diff
@@ -73,6 +73,53 @@ steps:
...
+      - docker build -t matrixdotorg/synapse:workers -f synapse-master/docker/Dockerfile-workers synapse-master/
+      # Build the accompanying Synapse workers Complement docker container. This specifically
+      # sets up a Synapse worker setup for Complement, and is located at:
+      # https://github.com/matrix-org/complement/blob/master/dockerfiles/SynapseWorkers.Dockerfile.
+      - docker build -t complement-synapse -f dockerfiles/SynapseWorkers.Dockerfile dockerfiles/
+      # Run the tests!
+      - COMPLEMENT_BASE_IMAGE=complement-synapse go test -v -tags synapse_blacklist ./tests
```
richvdh: this script could go in the complement repo, though it's not necessarily long enough to be worthwhile.

richvdh: why is `COMPLEMENT_BASE_IMAGE` set here, and a bunch of other `COMPLEMENT_*` vars are set in `environment`? consistency ftw.

complement/pipeline.yml
```diff
@@ -73,6 +73,53 @@ steps:
...
+    label: "\U0001F9EA Complement / Synapse Workers / :go: 1.15"
+    agents:
+      # Running every worker takes a beefy system
+      queue: "xlarge"
+    plugins:
+      - docker#v3.7.0:
+          # The dockerfile for this image is at https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile.
+          image: "matrixdotorg/complement:latest"
+          mount-buildkite-agent: false
+          # Complement needs to know if it is running under CI
```
richvdh: duplicate line
```suggestion
```

richvdh (changes requested):
I think abunch of

complement/pipeline.yml
anoadragon453: Unfortunately the dockerfile depends on additional files, named [docker/conf/worker.yaml.j2](https://github.com/matrix-org/synapse/blob/anoa/synapse_worker_docker/docker/conf/worker.yaml.j2) and [docker/configure_workers_and_start.py](https://github.com/matrix-org/synapse/blob/anoa/synapse_worker_docker/docker/configure_workers_and_start.py).

It's probably easier to download the whole repo instead of the three files individually (which may become more files in the future).

complement/pipeline.yml
anoadragon453: I'd like to keep it in here for now, so we have one place for the steps in the Complement pipeline and Synapse pipeline.

The `wget` and `tar` steps could probably be combined into one line as well. If it starts ballooning then moving it sounds more plausible.

complement/pipeline.yml
richvdh: fair enough.

Note that this incantation won't work until the dockerfile in question is present on synapse `master`.

complement/pipeline.yml
richvdh: > I'd like to keep it in here for now, so we have one place for the steps in the Complement pipeline and Synapse pipeline.

well, they're in separate pipeline files, so I'm not sure that's much of a reason. Still, it's fine for now.

complement/pipeline.yml
richvdh: On which note:

> I have confirmed that this pipeline succeeds locally.

are you *sure*?

complement/pipeline.yml
```diff
@@ -62,7 +62,57 @@ steps:
           mount-buildkite-agent: false
           # Complement needs to know if it is running under CI
```
richvdh: duplicate comment... I feel like I've said this before somewhere.

richvdh (commented):
> I think abunch of

what I meant to write here but apparently failed was that a bunch of the comments on #122 also apply here (also: vice versa, re `COMPLEMENT_BASE_IMAGE`).

I'd like to see a decision on https://github.com/matrix-org/complement/issues/65.

complement/pipeline.yml
anoadragon453: Ah, yes. I removed one instance, but there was another.

anoadragon453:
> what I meant to write here but apparently failed was that a bunch of the comments on #122 also apply here (also: vice versa, re `COMPLEMENT_BASE_IMAGE`).

Yeah, I've been applying some fixes to both sides as a result of your comments. It took me a while to figure out YAML anchors, but now that I'm a bit more confident in using them I'm trying to apply them to both pipelines.

> I'd like to see a decision on matrix-org/complement#65.

Noted, alright.

kegsay:
What's the status of this PR @anoadragon453 ?

anoadragon453:
@Kegsay blocked on https://github.com/matrix-org/synapse/pull/9162/, which is hopefully very nearly there at this point.

kegsay:
That has merged now, what's the status of this PR?

anoadragon453:
Blocked on Synapse Worker image not being flaky on startup: https://github.com/matrix-org/synapse/issues/10065

richvdh:
closing this, because complement has moved its CI to GHA.

PR #124, opened 2021-01-22
Add p35 old deps with postgres run for Synapse

erikjohnston:
Requires: https://github.com/matrix-org/synapse/pull/9206


richvdh (changes requested):
pending updates to https://github.com/matrix-org/synapse/pull/9206

ShadowJonathan:
Do we still want to test for py35 when it's been officially deprecated by synapse with the last version?

PR #125, opened 2021-01-22
Add pipeline for gomatrixserverlib

neilalexander:
Travis CI is really slow lately for some reason so figured we could use Buildkite.

### DIFF
--- a/gomatrixserverlib/pipeline.yml
+++ b/gomatrixserverlib/pipeline.yml
@@ -1,0 +1,21 @@
+steps:
+  - command:
+      - "env"
+      - "go test ./..."
+    label: "\U0001F9EA Unit tests / :go: 1.13"
+    plugins:
+      - docker#v3.0.1:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
+
+  - command:
+      - "env"
+      # https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
+      - "golangci-lint run"
+    label: "\U0001F9F9 Lint / :go: 1.13"
+    agents:
+      queue: "xlarge"
+    plugins:
+      - docker#v3.0.1:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
--- a/gomatrixserverlib/pipeline.yml
+++ b/gomatrixserverlib/pipeline.yml
@@ -1,6 +1,6 @@
 steps:
   - command:
-      - "env"
+      - "go version"
       - "go test ./..."
     label: "\U0001F9EA Unit tests / :go: 1.13"
     plugins:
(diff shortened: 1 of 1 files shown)


kegsay (dismissed):

gomatrixserverlib/pipeline.yml
```diff
@@ -0,0 +1,21 @@
+steps:
+  - command:
+      - "env"
```
richvdh: is this actually useful, or just noise?

gomatrixserverlib/pipeline.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "go test ./..."
+    label: "\U0001F9EA Unit tests / :go: 1.13"
+    plugins:
+      - docker#v3.0.1:
+          image: "golang:1.13"
+          mount-buildkite-agent: false
+
+  - command:
+      - "env"
+      # https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
```
richvdh: I'm not seeing how this link helps me?

gomatrixserverlib/pipeline.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "golangci-lint run"
+    label: "\U0001F9F9 Lint / :go: 1.13"
+    agents:
+      queue: "xlarge"
```
richvdh: REALLY?

richvdh (changes requested):

gomatrixserverlib/pipeline.yml
neilalexander: This is the size that's used by dendrite's linter. Are these sizes documented anywhere?

gomatrixserverlib/pipeline.yml
neilalexander: It does include the exact Go minor version of the image, but I suppose that can be replaced with `go version` if that's preferable.

gomatrixserverlib/pipeline.yml
neilalexander: Looking at the agents in BK it seems these are actually based on EC2 T3 sizes, so yes, 16GB does seem like a lot. `golangci-lint` is quite memory heavy but I'll put it down to `medium` (4GB RAM) for now.

kegsay (approved):

gomatrixserverlib/pipeline.yml
richvdh: supposedly the queues are documented at https://gitlab.matrix.org/new-vector/internal/-/wikis/BuildKite#build-queues, but they are a bit light on detail...

richvdh: (apart from saving us money, an advantage of using a smaller instance is that you are more likely to find a spare agent sitting around idle rather than having to wait while one gets spun up - or sitting in the queue, since we limit the number of xlarge agents more restrictively.)

richvdh (approved):

PR #126, opened 2021-01-25
Upgrade Element Web pipelines to modern Node and base OS

jryans:
This changes all Element Web steps to use Node 14 (current LTS) and Debian Buster (current stable) as the base OS.


richvdh (approved):
if you say so!

PR #130, opened 2021-01-30
Remove conditional for pypy builds

ShadowJonathan:


richvdh:
for the record: the thinking here is that this doesn't need to be conditional, given it's soft-fail.

richvdh (approved):

PR #134, opened 2021-02-12
Do not soft-fail Dendrite on sytest.

clokep:
To avoid situations like https://github.com/matrix-org/sytest/pull/1011#issuecomment-777867984

CC @neilalexander / @Kegsay

### DIFF
--- a/sytest/pipeline.yml
+++ b/sytest/pipeline.yml
@@ -214,8 +214,6 @@ steps:
       - matrix-org/annotate:
           path: "logs/annotate.md"
           style: "error"
-    soft_fail:
-      - exit_status: 1
     retry:
       automatic:
         - exit_status: -1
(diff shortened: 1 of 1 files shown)


clokep:
I'm not sure on the difference between the normal and "Full HTTP APIs" runs though this includes all of them.

kegsay:
Personally I only think we need 1 test (e.g Postgres/Monolith) given we're mainly checking for Sytest not for Dendrite. LGTM though

richvdh (changes requested):
I don't think we've ever had a clean run of these, so I don't think we can merge this yet.

See https://buildkite.com/matrix-dot-org/sytest/builds?branch=develop

clokep:
I'm going to close this since the ball isn't really in my court. 😢

PR #139, opened 2021-04-07
Enable Complement tests for MSC3083.

clokep:
Enable additional complement tests added in matrix-org/complement#98 for matrix-org/synapse#9735.

I think we need to merge the Synapse PR before these will pass?


richvdh (approved):
taking your word on this

PR #140, opened 2021-04-08
Update synapse pipeline, dropping Python 3.5 builds

richvdh:


synapse/pipeline.yml
```diff
@@ -400,7 +388,7 @@ steps:
       - "bash /bootstrap.sh synapse"
     plugins:
       - docker#v3.7.0:
-          image: "matrixdotorg/sytest-synapse:py38"
+          image: "matrixdotorg/sytest-synapse:bullseye"
```
richvdh, line 391: the py38 and py37 images are no longer maintained, so I'm updating them while I'm here.

richvdh:
There are altogether too many `9`s and `3`s in this diff, but I *think* it's right.

synapse/pipeline.yml
```diff
@@ -323,9 +308,12 @@ steps:
   # We mostly care about testing each topology.
   # To vary the Python and Postgres versions, we use Docker images which are based
   # on an assortment of Debian releases.
```
anoadragon453:
```suggestion
  # on an assortment of Linux distributions.
```

callahad (commented):
From looking at a previous pipeline run, I'd expect us to need to change:

- Trial (Old Deps) to Bionic
- Trial Py 3.5 to (removed)
- Trial Py 3.5 + Pg 9.5 to Py 3.6 + Pg 9.6
- PortDB Py 3.5 + Pg 9.5 to Py 3.6 + Pg 9.6

And then in SyTest, our 3 Stretch images (Monolith SQLite, Monolith PostgreSQL, Workers PostgreSQL) would need to change to... something (Bionic, again?)

synapse/pipeline.yml
```diff
@@ -323,9 +308,12 @@ steps:
...
-  #   - "py35" is Debian  9 Stretch (Python 3.5, Postgres 9.6)
-  #   - "py37" is Debian 10 Buster (Python 3.7, Postgres 11)
-  #   - "py38" is Debian 11 Bullseye (Python 3.9, Postgres 13)
+  #   - "bionic" (Ubuntu 18.04) has Python 3.6 and Postgres 10
+  #   - "buster" (Debian 10) has Python 3.7 and Postgres 11
+  #   - "bullseye" (Debian 11) (currently) has Python 3.9 and Postgres 13
+  #
+  # TODO: this leaves us without sytests for Postgres 9.6. How much do we care
+  #    about that?
```
erikjohnston, line 316: It's quite unfortunate, really. I wonder if we should be using docker compose here to allow us to more accurately select python and postgres versions? That's not really a small change though....

callahad:
This appears to change everything I would expect it to change.

callahad (dismissed):

synapse/pipeline.yml
callahad: This was discussed in #synapse-dev and the rough conclusion was that Pg9.6 goes EOL in November (only 7 months away! Eek!) so it's not worth addressing immediately since we do get coverage of the very next version by targeting Bionic.

And some time between now and November we'll do the proper thing and split out Python / Postgres versions. My hope is that if GitHub Actions proves fruitful making this sort of change will be (a) easier, and (b) accessible to the community (Jonathan has already volunteered).

erikjohnston (dismissed):

synapse/pipeline.yml
```diff
@@ -126,7 +126,8 @@ steps:
       TRIAL_FLAGS: "-j 2"
     plugins:
       - docker#v3.7.0:
```
anoadragon453, line 128: The label for this step needs to be updated to Python 3.6 as well.

anoadragon453 (commented):
Just some small things.

callahad:
I guess third review's the charm 😆 Nice catches, @anoadragon453

synapse/pipeline.yml
callahad: Some might claim that Ubuntu ontologically fits within the umbrella of "Ubuntu releases"

richvdh:
thank you all!

PR #141, opened 2021-04-09
Enable MSC2946 tests in Complement for Synapse and Complement pipelines.

clokep:
Pretty much the same as #139, but for matrix-org/synapse#9771 / matrix-org/complement#101.


richvdh (approved):
lgtm

PR #145, opened 2021-05-05
Fix Element Desktop lint step

jryans:
https://github.com/matrix-org/pipelines/pull/142 added an extra CI step, but since the plugins are part of each step, that change meant that the pre-existing lint step no longer had a Docker image.

This fixes things up by setting up the plugins for each step.

This is safe to merge at any time.


richvdh (approved):

richvdh:
took me a while to clock you were talking about buildkite plugins rather than lint plugins :laughing:

PR #149, opened 2021-05-20
First compile, then run the lint check.

bmarty:


richvdh (approved):
looks fine, but I'm curious about why. most people like to run the lint step first as it's usually quicker.

PR #150, opened 2021-05-22
Fix syntax error in dendrite/pipeline.yml

thaumavorio:
A small error currently prevents the YAML from parsing.


anoadragon453 (approved):
Good spot, thank you!

PR #153, opened 2021-05-27
Use "npm ci" to install dependencies for IRC bridge

andir:
This has the benefit of ensuring that the lockfile in the repository is
in the correct state as `npm ci` only installs but doesn't update the
lockfile. It is otherwise very simple to miss updating the lockfile when
changing dependencies as (the yarn invocation) would just happily update
the lockfile on every invocation regardless of the state of the lockfile
in the repository.

The two locking mechanisms of NPM and Yarn are not compatible but both
are viable these days. NPM has the benefit of not requiring additional
tooling (besides the node toolchain) and being compatible with
downstream packing tooling (that consumes those lockfiles).

The CI task has so far already spammed something along the lines of:

>  No lockfile found.
> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

### DIFF
--- a/matrix-appservice-irc/pipeline.yml
+++ b/matrix-appservice-irc/pipeline.yml
@@ -1,8 +1,8 @@
 steps:
   - label: ":eslint: Lint"
     command:
-      - "yarn"
-      - "yarn lint"
+      - "npm ci"
+      - "npm run-script lint"
     plugins:
       - docker#v3.0.1:
           image: "node:14"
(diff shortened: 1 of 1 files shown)


richvdh:
No objections from the CI side, but needs a review from someone in IRC land

richvdh (approved):

Half-Shot (approved):
No objections my side, if it works. In the past we had issues with npm doing things with paths that it shouldn't be able to access, which is why we used yarn.

PR #157, opened 2021-06-08
Pull the corresponding complement branch in the Synapse pipeline.

clokep:
This is similar to what we do in sytest where we run the CI against a matching branch name, if one exists. (Otherwise we fallback to `master`.)

We'll probably want similar changes on the complement side.


clokep:
I ran this locally and it seemed OK.

clokep:
CC @anoadragon453 and @kegsay who I've talked with about this recently.

anoadragon453 (approved):
Looks good from a cursory review.

kegsay (approved):
Looks great, we should do the same for Dendrite!

PR #158, opened 2021-06-09
Build Synapse from develop (or same named branch) instead of latest for complement.

clokep:
Currently the Synapse image used is from `synapse:latest` on DockerHub. This makes it annoying to develop a feature and at the same time update the tests in complement.

This is pretty much the same as #157, but from the complement pipeline side

### DIFF
--- a/complement/pipeline.yml
+++ b/complement/pipeline.yml
@@ -23,8 +23,13 @@ steps:
           limit: 3
 
   - command:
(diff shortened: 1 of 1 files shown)


clokep:
@anoadragon453 Would appreciate some 👀 on this too!

kegsay (approved):

anoadragon453 (approved):
Looks correct to me!

PR #159, opened 2021-06-09
Run knocking Complement tests in Synapse and Complement pipelines

anoadragon453:
This PR runs the Complement tests for [MSC2403](https://github.com/matrix-org/matrix-doc/pull/2403) (knocking) on the Synapse and Complement pipelines.

Requires https://github.com/matrix-org/pipelines/pull/158, as we'll need the Complement tests to test against Synapse's `develop` branch (which includes support for knocking).


clokep (approved):
Seems like the right change! 🎉

kegsay (approved):

PR #160, opened 2021-07-01
Remove references to complement's `master` branch

richvdh:
I'd like us to consider renaming to `main`, so let's remove the references to `master` specifically.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -38,6 +38,7 @@ steps:
 
   - command:
       - docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile .
+      - "mkdir -p complement"
       - "wget -O - https://github.com/matrix-org/complement/archive/HEAD.tar.gz | tar -xz --strip-components=1 -C complement"
       - "cd complement && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v -tags=\"dendrite_blacklist\" ./tests"
     label: "\U0001F9EA Complement"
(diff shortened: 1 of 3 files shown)


dendrite/pipeline.yml
```diff
@@ -38,14 +38,14 @@ steps:
 
   - command:
       - docker build -t complement-dendrite -f build/scripts/Complement.Dockerfile .
-      - "wget -N https://github.com/matrix-org/complement/archive/master.tar.gz && tar -xzf master.tar.gz"
-      - "cd complement-master && COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -v -tags=\"dendrite_blacklist\" ./tests"
+      - "wget -O - https://github.com/matrix-org/complement/archive/HEAD.tar.gz | tar -xz --strip-components=1 -C complement"
```
richvdh, line 42: the `--strip-components=1 -C complement` means that we don't care about whether the folder in the tarball is named `complement-master` or named after whatever `HEAD` happens to point to right now.

synapse/pipeline.yml
```diff
@@ -234,8 +234,8 @@ steps:
     plugins:
       - matrix-org/download#v1.1.0:
           urls:
-            - https://raw.githubusercontent.com/matrix-org/pipelines/master/synapse/docker-compose.yaml
-            - https://raw.githubusercontent.com/matrix-org/pipelines/master/synapse/docker-compose-env
+            - https://raw.githubusercontent.com/matrix-org/pipelines/HEAD/synapse/docker-compose.yaml
```
richvdh, line 237: thought I might as well do these while I was here.

dendrite/pipeline.yml
anoadragon453: FYI you'll need to `mkdir complement` as `tar` expects the directory to exist before it writes into it.

anoadragon453 (dismissed):
lgtm besides the one change below. TIL one can just pop `HEAD` in there though!

PR #169, opened 2021-07-22
Update complement invocation

kegsay:


anoadragon453 (approved):
Note that this is off the back of https://github.com/matrix-org/complement/pull/171.

PR #174, opened 2021-09-07
Swap out js sdk release-mode script for js script

dbkr:
The buildkite builders don't have jq so this never worked.
Just make it a js script instead.

Requires https://github.com/matrix-org/matrix-react-sdk/pull/6759

### DIFF
--- a/matrix-react-sdk/pipeline.yaml
+++ b/matrix-react-sdk/pipeline.yaml
@@ -18,7 +18,11 @@ steps:
       - "echo '+++ Lint'"
       - "yarn lint:types"
       - "echo '--- Switch js-sdk to release mode'"
-      - "./scripts/ci/js-sdk-to-release.sh"
+      - "./scripts/ci/js-sdk-to-release.js"
+      - "pushd node_modules/matrix-js-sdk"
+      - "yarn run build:compile"
+      - "yarn run build:types"
+      - "popd"
       - "echo '+++ Lint (release mode)'"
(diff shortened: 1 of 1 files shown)


matrix-react-sdk/pipeline.yaml
```diff
@@ -18,7 +18,11 @@ steps:
       - "echo '+++ Lint'"
       - "yarn lint:types"
       - "echo '--- Switch js-sdk to release mode'"
-      - "./scripts/ci/js-sdk-to-release.sh"
+      - "./scripts/ci/js-sdk-to-release.js"
+      - "pushd node_modules/matrix-js-sdk"
+      - "yarn run build:compile"
+      - "yarn run build:types"
+      - "popd"
```
richvdh, line 25: wouldn't it be more appropriate to just change `scripts/ci/js-sdk-to-release.sh` (or make another script)?

Tbh this whole thing (from `install-deps` through to `yarn lint:types` should be in a script in the repo. Having 59 steps in the build pipeline increases the coupling between the `matrix-react-sdk` repo and the `pipelines` repo, making it harder to change things in future (because both repos have to updated at once).

richvdh (changes requested):

dbkr:
Yeah, fair - in fact what this has made me realise is that I really ought to just spend the 30 mins required to move this to GHA, so I did: https://github.com/matrix-org/pipelines/pull/175

PR #177, opened 2021-11-23
Set `SYTEST_BRANCH` for Dendrite

neilalexander:
If this works then I'll revert matrix-org/sytest@23811a4 which doesn't appear to do the right thing anymore anyway.


kegsay (approved):
What does Synapse do for this though?

neilalexander:
Apparently they do something similar but in GHA: https://github.com/matrix-org/synapse/blob/6a5dd485bd82b269e7e169c0385290d081eae801/.github/workflows/tests.yml#L204

PR #178, opened 2021-12-03
Add Windows build step for Dendrite

S7evinK:
As the title says, this adds a Windows build step for Dendrite.
This will catch issues like matrix-org/dendrite/issues/1980


neilalexander (approved):
Seems like a good idea!

PR #179, opened 2021-12-30
Remove the Synapse LDAP3 BuildKite configuration as it's obsolete

reivilibre:
Migrating to GitHub Actions.
See https://github.com/matrix-org/matrix-synapse-ldap3/pull/145.


richvdh (approved):
:tada:

PR #180, opened 2022-01-10
Move matrix-appservice-bridge from node 12,14 to 14,16

Half-Shot:


richvdh (approved):
ok. Please could you consider porting this to GHA though?

PR #181, opened 2022-01-18
Move Mjolnir from node 12 to node 14

Gnuxie:
See https://github.com/matrix-org/mjolnir/pull/192.


richvdh (approved):
sure, but it would be good to migrate this to GHA.

PR #182, opened 2022-01-18
Update rageshake build to use golang v1.16 / v1.17.

michaelkaye:
These are the current stable builds, and the lint step does
not seem to pass with go v1.15.

https://buildkite.com/matrix-dot-org/rageshake/builds/42#000a370f-9c12-427a-b3ae-2ef14bec30be for an example of the failure; local testing shows that running the linter using go v1.15 fails with the same error; 1.16/1.17 are fine.

Assuming it's just backwards incompatibility on go's front here.

### DIFF
--- a/rageshake/pipeline.yml
+++ b/rageshake/pipeline.yml
@@ -12,14 +12,7 @@ steps:
     command: ".buildkite/lint.sh"
     plugins:
       - docker#v3.0.1:
-          image: "golang:1.15"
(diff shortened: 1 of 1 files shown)


richvdh (approved):
sure, though we should really move this to GHA.

richvdh:
(also, we'd better make sure we're testing the version we use in production)

PR #185, opened 2022-01-20
Propagate environment variables for Windows build

S7evinK:
The last runs didn't create Windows binaries, since the environment wasn't used.

### DIFF
--- a/dendrite/pipeline.yml
+++ b/dendrite/pipeline.yml
@@ -25,6 +25,7 @@ steps:
       - docker#v3.0.1:
           image: "golang:1.15"
           mount-buildkite-agent: false
+          propagate-environment: true
     retry:
       automatic:
         - exit_status: 128


kegsay:
Not sure I understand: why do we need to `propagate-environment` on Windows?

S7evinK:
The environment vars aren't "global", so they didn't make it to the docker plugin, unfortunately.
I can move the env vars to the docker plugin, if that's preferred.

kegsay (approved):

dendrite/pipeline.yml
```diff
@@ -25,6 +25,7 @@ steps:
       - docker#v3.0.1:
           image: "golang:1.15"
           mount-buildkite-agent: false
+          propagate-environment: true
```
richvdh, line 28: maybe too late now, but some comments as to why this is needed might have been been useful. (Or, yes, moving the `env` settings onto the docker plugin)

S7evinK: Since `propagate-environment` is used in other steps, I didn't think that was necessary.
[Docs](https://github.com/buildkite-plugins/docker-buildkite-plugin/tree/v3.9.0#propagate-environment-optional-boolean) for this. :)

richvdh: mmm, fair.

PR #191, opened 2022-03-04
Exodus changed their docker entrypoint in a minor release.

michaelkaye:
Previously the entrypoint included app.apk; now the entrypoint is just the script, and the APK is passed via the CMD.

The `-j` flag we passed is overwriting the CMD - so we need to add the app.apk back in to let the changed docker container work.

Fixes the build being broken in vector-im/element-android#5422 (but a TODO is still to migrate to GHA)

PR #192, opened 2022-03-04
Fixing android exodus usage

ouchadam:
The latest version of exodus (which automatically gets pulled in) updates the docker entry points, breaking our existing usage

- Fixed by splitting the directory from the apk and providing the apk as a dedicated param

*running locally 

| BEFORE | AFTER |
| --- | --- |
| 2022-03-04T11:10:39,717996680+00:00|2022-03-04T11:22:57,274156579+00:00

Their readme was also updated to reflect the change https://github.com/Exodus-Privacy/exodus-standalone/blob/master/README.md#using-docker

### DIFF
--- a/element-android/pipeline.yml
+++ b/element-android/pipeline.yml
@@ -70,7 +70,7 @@ steps:
 
     commands:
       - "mv vector-gplay-universal-*.apk app.apk"
-      - "docker run -v $(pwd):/app.apk --rm -i exodusprivacy/exodus-standalone /app/app.apk -j > exodus.json"
+      - "docker run -v $(pwd):/app --rm -i exodusprivacy/exodus-standalone /app/app.apk -j > exodus.json"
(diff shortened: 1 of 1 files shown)


bmarty (approved):
LGTM, @richvdh can you merge it please?

ariskotsomitopoulos (approved):
👍

ouchadam:
going to close in favour of #191, wasn't quite fast enough :racehorse:

element-android/pipeline.yml
```diff
@@ -70,7 +70,7 @@ steps:
 
     commands:
       - "mv vector-gplay-universal-*.apk app.apk"
-      - "docker run -v $(pwd)/app.apk:/app.apk --rm -i exodusprivacy/exodus-standalone -j > exodus.json"
```
michaelkaye, line 73: This change technically mounts more than we need into the exodus container; i prefer without the /app mount as it's not required.

It is what the exodus docs suggest though, so *shrug*

PR #194, opened 2022-03-28
Remove reskindex from react-sdk pipeline

turt2live:
**Blocked on https://github.com/matrix-org/matrix-react-sdk/pull/7962**


michaelkaye (approved):
We've removed this command from the package.json so we can't run it if we wanted to (after the matrix-react-sdk change merges).

Good to go once linked PR merges.

PR #197, opened 2022-07-06
Remove Mjolnir

Gnuxie:
We have migrated to github actions https://github.com/matrix-org/mjolnir/pull/316.


richvdh (approved):
Thanks. I've also archived the `mjolnir` pipeline.

PR #198, opened 2022-07-15
Drop matrix-appservice-slack from pipelines

Half-Shot:
We won't be using this once https://github.com/matrix-org/matrix-appservice-slack/pull/688 is merged. The pipeline on BK can also be deleted once this is merged.


richvdh (approved):
thanks. I've also archived the pipeline in buildkite.

PR #199, opened 2022-07-20
Drop matrix-appservice-node from pipelines

Half-Shot:
Depends on https://github.com/matrix-org/matrix-appservice-node/pull/57. You can also archive the buildkite pipeline after this.


richvdh (approved):
lgtm

richvdh:
> You can also archive the buildkite pipeline after this.

done

PR #205, opened 2022-10-28
Remove unused ios pipelines

benbz:
cc @manuroe @stefanceriu 

The following have been archived in BuildKite in the past but not cleared up here:
* https://buildkite.com/matrix-dot-org/element-ios
* https://buildkite.com/matrix-dot-org/matrix-ios-kit
* https://buildkite.com/matrix-dot-org/matrix-ios-sdk

Cleaning up now whilst looking at others


manuroe (approved):
Thanks a lot for doing this cleanup!

PR #209, opened 2022-10-28
Remove optimized-python-docker, unused

benbz:
Looks unused. Can't even find an archived pipeline on BuildKite for this


richvdh (approved):
no idea what that was about

